PR #14, opened 2022-06-07
Apply external ignored_labels to C-IL for detection

harimkang:
### This PR includes :
- Incoming ignored_labels from OTE is received from 'img_metas' to create an ignored mask
- Deliver ignored masks to loss to use as a mask (Modify the forward & loss function on the head side to receive the ignored_mask)


mpa/modules/models/heads/cross_dataset_detector_head.py
```diff
@@ -236,4 +217,15 @@ def vfnet_to_atss_targets(self,
...
-        return labels_list, label_weights, bbox_targets_list, bbox_weights
+        return labels_list, label_weights, bbox_targets_list, bbox_weights, ignored_masks
+
+    def get_ignored_masks(self, img_metas, all_labels):
+        ignored_masks = []
+        for i, meta in enumerate(img_metas):
+            mask = torch.Tensor([1 for _ in range(self.num_classes)])
+            if 'ignored_labels' in meta and meta['ignored_labels']:
+                mask[meta['ignored_labels']] = 0
+            mask = mask.repeat(len(all_labels[i]), 1)
```
chuneuny-emily, line 228: Do you need mask.repeat code? What is the purpose?

harimkang: @chuneuny-emily Yes, this is a function that takes ignored_labels and makes them a mask that will be used for loss. Because information about which images each label belongs to is not passed to loss, the ignored_labels are replaced with masks in the head.

chuneuny-emily: > @chuneuny-emily Yes, this is a function that takes ignored_labels and makes them a mask that will be used for loss. Because information about which images each label belongs to is not passed to loss, the ignored_labels are replaced with masks in the head.

I understood.

chuneuny-emily (approved):
It looks good for me.

mpa/modules/models/heads/custom_vfnet_head.py
```diff
@@ -154,19 +160,22 @@ def loss(self,
...
             if isinstance(self.loss_cls, CrossSigmoidFocalLoss):
                 loss_cls = self.loss_cls(
                     flatten_cls_scores,
                     cls_iou_targets,
-                    weight=label_weights.unsqueeze(-1),
+                    weight=label_weights,
                     avg_factor=num_pos_avg_per_gpu,
                     use_vfl=self.use_vfl,
-                    use_weight=self.use_atss)
+                    use_weight=self.use_atss,
```
goodsong81: Could you explain why?

goodsong81 (approved):
LGTM. Could you answer the question in the comment?

mpa/modules/models/heads/custom_vfnet_head.py
harimkang: I missed that part and didn't push it. I removed it.

mpa/modules/models/heads/cross_dataset_detector_head.py
```diff
@@ -236,4 +217,15 @@ def vfnet_to_atss_targets(self,
         ]
         label_weights = torch.cat(label_weights_list)
         bbox_weights = torch.cat(bbox_weights_list)
-        return labels_list, label_weights, bbox_targets_list, bbox_weights
+        return labels_list, label_weights, bbox_targets_list, bbox_weights, ignored_masks
+
+    def get_ignored_masks(self, img_metas, all_labels):
```
goodsong81: Is this 'ignored' mask or 'valid' mask?
Actually, valid_label_mask would be better in terms of clarification.

goodsong81 (changes requested):

mpa/modules/models/heads/cross_dataset_detector_head.py
harimkang: The ignored_masks were modified to valid_label_mask.

goodsong81 (approved):

PR #15, opened 2022-06-08
Integrate C-IL for multi-label classification

JihwanEom:
### This PR includes :
- Integrate class-incremental learning for multi-label classification
- Improve stability in stage.py


mpa/cls/stage.py
```diff
@@ -74,7 +74,10 @@ def configure(self, model_cfg, model_ckpt, data_cfg, training=True, **kwargs):
                 cfg.data.num_classes = len(cfg.data.train.get('classes', []))
             cfg.model.head.num_classes = cfg.data.num_classes
 
-        if isinstance(cfg.model.head.topk, tuple):
+        if 'Multi' in cfg.model.head.get('type', False) and cfg.model.head.get('topk', False):
```
harimkang: Can't we use cfg.task as the standard (not head.type)?
I think we can also combine this if statement with 79-80 line and use it as one if statement.

models/classification/ote_efficientnet_b0_multilabel.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+model:
+    type: SAMImageClassifier
+    task: classification
+    backbone:
+        version: b0
+    head:
+        type: MultiLabelLinearClsHead
+        loss:
+            type: AsymmetricLoss
```
harimkang: need to add new line

mpa/cls/stage.py
```diff
@@ -196,12 +199,14 @@ def configure_task(cfg, training, model_meta=None, **kwargs):
...
                 cfg.model.head.num_classes = len(dst_classes)
                 gamma = 2 if cfg['task_adapt'].get('efficient_mode', False) else 3
-                cfg.model.head.loss = ConfigDict(
-                    type='SoftmaxFocalLoss',
-                    loss_weight=1.0,
-                    gamma=gamma,
-                    reduction='none',
-                )
+
+                if 'Multi' not in cfg.model.head.type:
```
harimkang, line 203: As in the previous comment, we can use (head.type ->) cfg.task as the standard task type.

harimkang (changes requested):

mpa/cls/stage.py
harimkang: sorry, I was mistaken. This is using the same cfg.task as 'classification'.
I think you can change cfg.task to multi-label classification from classification 
or The conditional statement {'Multi' in cfg.model.head.get('type') }needs to be reconsidered. (need to be more specific.)

mpa/cls/stage.py
harimkang: Same as above comment.

goodsong81 (approved):
LGTM. Please resolve Harim's comments.

mpa/cls/stage.py
```diff
@@ -196,12 +199,14 @@ def configure_task(cfg, training, model_meta=None, **kwargs):
...
                 cfg.model.head.num_classes = len(dst_classes)
                 gamma = 2 if cfg['task_adapt'].get('efficient_mode', True) else 3
-                cfg.model.head.loss = ConfigDict(
-                    type='SoftmaxFocalLoss',
-                    loss_weight=1.0,
-                    gamma=gamma,
-                    reduction='none',
-                )
+
+                if not cfg.model.get('multilabel', False):
```
harimkang, line 201: 'If not' doesn't look good, do you have any other ideas?

mpa/cls/stage.py
```diff
@@ -74,8 +74,11 @@ def configure(self, model_cfg, model_ckpt, data_cfg, training=True, **kwargs):
                 cfg.data.num_classes = len(cfg.data.train.get('classes', []))
             cfg.model.head.num_classes = cfg.data.num_classes
 
-        if isinstance(cfg.model.head.topk, tuple):
-            cfg.model.head.topk = (1,) if cfg.model.head.num_classes < 5 else (1, 5)
+        if cfg.model.head.get('topk', False) and isinstance(cfg.model.head.topk, tuple):
+            if cfg.model.get('multilabel', False):
```
harimkang, line 79: how about
```
cfg.model.head.topk = (1,) if cfg.model.head.num_classes < 5 else (1, 5)
if cfg.model.get('multilabel', False): cfg.model.head.pop('topk', None)
```

harimkang (changes requested):
I left a few comments.

mpa/cls/stage.py
JihwanEom: How do you think about below code?

```
cfg.model.head.topk = (1, cfg.model.head.num_classes) if cfg.model.head.num_classes < 5 else (1, 5)
if cfg.model.get('multilabel', False): cfg.model.head.pop('topk', None)
```

mpa/modules/models/classifiers/sam_classifier.py
```diff
@@ -30,6 +30,8 @@ class SAMImageClassifier(ImageClassifier):
     """SAM-enabled ImageClassifier"""
 
     def __init__(self, task_adapt=None, **kwargs):
+        if 'multilabel' in kwargs:
```
JihwanEom, line 33: @harimkang this multilabel flag is needed only for state_dict_hook. How do you think about separating state_dict_hook into two cases: 1) backbone, 2) head
If we separate this, each classifier will not have to remove multilabel argument. But I think state_dict hook in heads might be implemented.

harimkang (approved):
LGTM.
Need to check that there are no conflicts with recently merged code.

supersoob (approved):
LGTM

PR #16, opened 2022-06-09
Apply to use external ignored_labels to C-IL for semantic segmentation

chuneuny-emily:
### This PR includes :
- Apply to use external ignored_labels from OTE
- Add custom fcn/ocr heads to create ignored mask based on external ignored_labels from 'img_metas'
- Loss masking based on the external ignored_labels


mpa/modules/models/heads/custom_fcn_head.py
```diff
@@ -0,0 +1,123 @@
...
+from mmseg.ops import resize
+from mmseg.core import add_prefix
+from mpa.modules.utils.mask_utils import get_ignored_labels_per_batch
+
+@HEADS.register_module()
+class CustomFCNHead(FCNHead):
+    """Custom Fully Convolution Networks for Semantic Segmentation.
+    """
+
+    def __init__(self,
```
harimkang: As I looked at it, FCNHead inherits the BaseDecodeHead.  The __init__ contents of CustomFCNHead match the init contents of BaseDecodeHead, is there a reason why you overridding this function? (I think it will work well same without the __init__).

harimkang (commented):
I've looked at it a little bit. I'll take the time to look at it in more detail later.

mpa/modules/models/heads/custom_fcn_head.py
chuneuny-emily: You're right. I missed to remove the duplicated codes. I'll remove the contents in init. Thanks!

mpa/modules/utils/mask_utils.py
```diff
@@ -0,0 +1,14 @@
...
+import torch
+
+def get_ignored_labels_per_batch(img_metas, num_classes):
+    ignored_labels = []
+    for _, meta in enumerate(img_metas):
+        valid_labels = torch.Tensor([1 for _ in range(num_classes)])
+        if 'ignored_labels' in meta and meta['ignored_labels']:
+            valid_labels[meta['ignored_labels']] = 0
+        ignored_labels.append(valid_labels)
+    return ignored_labels
```
goodsong81: mask_utils.py <- this name could be misleading. Is this a utility function for mask?

goodsong81: Is this function getting 'valid' labels per batch? Or 'ignored' labels per batch?
@harimkang What do you think? Name and contents are not consistent.

mpa/modules/models/heads/custom_fcn_head.py
goodsong81: Would you remove __init__() in this PR?

goodsong81 (changes requested):
Maybe we need to refine some variable/module names.
Ignored? Valid? Mask?

mpa/modules/utils/mask_utils.py
chuneuny-emily: It's getting 'valid' labels per batch. Let me rename it to  get_valid_labels_per_batch.

chuneuny-emily: > 
mask_utils.py was created to put common function named get_valid_labels_per_batch, which are called in custom fcn/ocr heads. What do you think of renaming it to seg_utils.py?

harimkang: I will modify the detection side to the same name ('ignored' -> 'valid') 
and the name of the function that gets the mask to 'get_valid_labels_per_batch'.

chuneuny-emily: I had modified it to 'valid' yesterday. But valid_labels name has been already using in mmseg. So I need to modify the name again. (We cannot name it to valid_labels or valid_mask because they have been using in mmseg.) 

What do you think of keeping using ignored_labels or ignored_mask?

goodsong81: How about valid_label_mask? This seems more precisely correct for this code semantic. If you want to keep it as ignored_xxx, we need to flip 0/1 values.
valid = [0, 0, 1, 0] -> ignored = [1, 1, 0, 1]

harimkang: The valid_label_mask looks good, too.

chuneuny-emily: Looks good. Let me rename it to valid_label_mask. Thanks for your suggestion.

mpa/modules/models/heads/custom_fcn_head.py
```diff
@@ -2,59 +2,29 @@
...
-
     def forward_train(self, inputs, img_metas, gt_semantic_seg, train_cfg, pixel_weights=None):
         """Forward function for training.
         Args:
             inputs (list[Tensor]): List of multi-level img features.
             img_metas (list[dict]): List of image info dict where each dict
                 has: 'img_shape', 'scale_factor', 'flip', and may also contain
                 'filename', 'ori_shape', 'pad_shape', 'img_norm_cfg',
-                and 'ignored_labels'.
+                and 'valid_labels'.
```
goodsong81: Maybe 'ignored_labels' is correct for this part.

mpa/modules/models/heads/custom_ocr_head.py
```diff
@@ -27,7 +25,7 @@ def forward_train(self, inputs, prev_output, img_metas, gt_semantic_seg, train_c
             img_metas (list[dict]): List of image info dict where each dict
                 has: 'img_shape', 'scale_factor', 'flip', and may also contain
                 'filename', 'ori_shape', 'pad_shape', 'img_norm_cfg',
-                and 'ignored_labels'.
+                and 'valid_labels'.
```
goodsong81: Same here. We get 'ignored_labels' from the img_metas. Then we convert it to 'valid_label_mask'.

goodsong81 (changes requested):

mpa/modules/models/heads/custom_fcn_head.py
chuneuny-emily: fixed

mpa/modules/models/heads/custom_ocr_head.py
chuneuny-emily: fixed

goodsong81 (approved):
LGTM.

PR #19, opened 2022-06-20
Integrate C-IL for instance segmentation

harimkang:
### This PR includes :
- Integrate class-incremental learning for instance segmentation task (recipe)
- Add model template {EfficientNetB2B-maskRCNN & Resnet50-maskRCNN}
- Add ignore mode for instance-segmentation
- Modify the code of the infer function to be the same as OTE detection inference_task
- related OTE PR: https://github.com/openvinotoolkit/training_extensions/pull/1142


mpa/det/stage.py
```diff
@@ -165,8 +152,9 @@ def configure_task_classes(self, cfg, task_adapt_type, task_adapt_op):
         # Model classes
         if task_adapt_op == 'REPLACE':
             if len(data_classes) == 0:
-                raise ValueError('Data classes should contain at least one class!')
-            model_classes = data_classes.copy()
+                model_classes = org_model_classes.copy()
+            else:
+                model_classes = data_classes.copy()
         elif task_adapt_op == 'MERGE':
```
harimkang, line 158: @goodsong81 @JihwanEom @jaegukhyun @chuneuny-emily @supersoob 
Hi, I share this because I have a problem to think about together. 

The current default value for task_adapt_op is 'MERGE'. It is done in most tasks, and it is not a problem in C-IL setting. The problem is that sometimes org_model_classes are imported when using 'coco pretrained weights' or other pretrained weights in a 'Supervised Learning setting'. (ex, 80 classes for coco) 

To solve this problem, we can fix the code as above (155-157 lines) with 'REPACE' as default. This is the problem that appears in (the instance-segmentation ResNet pretrained weight) and detection (Yolox pretrained weight) so far.

If you have any concerns or ideas, please share them. Also, I wonder if this happens in other tasks.

goodsong81 (approved):
LGTM.

mpa/det/stage.py
goodsong81: No objection for 'REPLACE' default. In SC/OTE the data_classes is the final target classes anyway.

PR #20, opened 2022-06-30
Pretrained weights bugfix

JihwanEom:
### This PR includes :
- Set pretrained=True as default in efficientnet, mobilenet backbones
- Bugfix for init_weights method.

### DIFF
--- a/models/classification/_base_/ote_efficientnet.py
+++ b/models/classification/_base_/ote_efficientnet.py
@@ -3,6 +3,7 @@ model = dict(
     type='ImageClassifier',
     backbone=dict(
         type='OTEEfficientNet',
+        pretrained=True,
         version='b0'),
     neck=dict(type='GlobalAveragePooling'),
     head=dict(
--- a/mpa/cls/stage.py
+++ b/mpa/cls/stage.py
@@ -38,11 +38,12 @@ class ClsStage(Stage):
             raise ValueError(
                 f'Given model_cfg ({model_cfg.filename}) is not supported by classification recipe'
             )
-        self.configure_model(cfg, training, **kwargs)
-
+            
         # Checkpoint
         if model_ckpt:
             cfg.load_from = self.get_model_ckpt(model_ckpt)
+
+        self.configure_model(cfg, training, **kwargs)
 
         # OMZ-plugin
         if cfg.model.backbone.type == 'OmzBackboneCls':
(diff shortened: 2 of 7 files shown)


models/classification/_base_/ote_efficientnet.py
```diff
@@ -3,6 +3,7 @@
     type='ImageClassifier',
     backbone=dict(
         type='OTEEfficientNet',
+        pretrained=True,
```
harimkang, line 6: If this is all true, doesn't C-IL have a problem? or does C-IL put it in the model as false?

JihwanEom: The pretrained weight is loaded first, and then overwritten with C-IL weight. I've checked smart weight loading also works fine without any error. Thank you for your opinion.

supersoob (commented):
Please update efficientnet_v2_s. It's a default model for SC.

JihwanEom:
EffNet v2 doesn't need any update about this issue. effnet v2 uses pretrained weights as default.

Please refer here: https://github.com/openvinotoolkit/model_preparation_algorithm/blob/develop/mpa/modules/models/backbones/efficientnetv2.py#L46

mpa/cls/stage.py
```diff
@@ -89,6 +90,10 @@ def configure(self, model_cfg, model_ckpt, data_cfg, training=True, **kwargs):
     def configure_model(cfg, training, **kwargs):
         # verify and update model configurations
         # check whether in/out of the model layers require updating
+
+        if cfg.get('load_from', None) and cfg.model.backbone.get('pretrained', None):
```
JihwanEom, line 94: @harimkang How do you think about this order change? I changed to get cfg.load_from variable in configure_model. I couldn't find any error for both multi-class/multi-label tasks. Is there any problem that I missed you think?

supersoob, line 94: Why don't you make it False rather than None

JihwanEom: @supersoob @harimkang How do you think about this order change? I changed to get cfg.load_from variable in configure_model. I couldn't find any error for both multi-class/multi-label tasks. Is there any problem that I missed you think?

supersoob: Could you make disable pretrained in eff_v2? I think its pretrained model will be loaded eventually even you set it as None.

supersoob (approved):
LGTM

PR #21, opened 2022-07-01
Ignore mode for classification bugfix

JihwanEom:
### This PR includes:
- ignore mode for multi-label classification bugfix (only subset of ignored dataset was applied ignore mode -> all ignored dataset)

### DIFF
--- a/mpa/modules/models/classifiers/sam_classifier.py
+++ b/mpa/modules/models/classifiers/sam_classifier.py
@@ -78,7 +78,8 @@ class SAMImageClassifier(ImageClassifier):
         x = self.extract_feat(img)
 
         losses = dict()
-        if kwargs.get('img_metas', False) and len(kwargs['img_metas'][-1]['ignored_labels']) > 0:
+
+        if self.multilabel:
             loss = self.head.forward_train(x, gt_label, **kwargs)
         else:
             loss = self.head.forward_train(x, gt_label)


JihwanEom:
Passed All API/CLI test

harimkang (approved):

mpa/modules/models/classifiers/sam_classifier.py
```diff
@@ -78,7 +78,8 @@ def forward_train(self, img, gt_label, **kwargs):
         x = self.extract_feat(img)
 
         losses = dict()
-        if kwargs.get('img_metas', False) and len(kwargs['img_metas'][-1]['ignored_labels']) > 0:
+
+        if self.multilabel:
             loss = self.head.forward_train(x, gt_label, **kwargs)
         else:
             loss = self.head.forward_train(x, gt_label)
```
goodsong81, line 85: Why do we need this condition by the way? The prototype allows extra params anyway.
image

JihwanEom: In mpa, multi-class heads don't allow extra params.
https://github.com/open-mmlab/mmclassification/blob/v0.9.0/mmcls/models/heads/linear_head.py#L53
https://github.com/openvinotoolkit/model_preparation_algorithm/blob/develop/mpa/modules/models/heads/non_linear_cls_head.py#L85

But above 0.18.0 version for mmcls, linear_head also allows extra params.
https://github.com/open-mmlab/mmclassification/blob/v0.18.0/mmcls/models/heads/linear_head.py#L49

I think this condition can be removed if we upgrade mmcls version.

PR #22, opened 2022-07-04
Tuning training configs for SC integration

harimkang:
### This PR includes :
- Model & training schedule Config value update
- Related OTE PR: https://github.com/openvinotoolkit/training_extensions/pull/1150
- Prevent [data_cls] -> [data_cls] training when run without previous label schema in classification.


mpa/cls/stage.py
```diff
@@ -206,20 +202,23 @@ def configure_task(cfg, training, model_meta=None, **kwargs):
...
-                    )
                     sampler_type = 'balanced'
+                    if len(model_classes) > 0:
+                        cfg.model.head.loss = ConfigDict(
+                            type='SoftmaxFocalLoss',
+                            loss_weight=1.0,
+                            gamma=gamma,
+                            reduction='none',
+                        )
+                    elif len(model_classes) == 0:
```
harimkang: @supersoob @JihwanEom  I think it's best to just use 'else'.

harimkang: Also, I don't think model_classes will always be zero for the Supervised Learning setting, what do you think?
For active learning and supervised learning, model_classes is expected to be greater than zero.

mpa/modules/hooks/early_stopping_hook.py
```diff
@@ -62,12 +60,14 @@ def __init__(self,
                  rule: Optional[str] = None,
                  patience: int = 5,
                  iteration_patience: int = 500,
-                 min_delta: float = 0.0):
+                 min_delta: float = 0.0,
+                 warmup_iters: int = 0):
```
harimkang: @chuneuny-emily Is there a background in which this was added?

chuneuny-emily: It was required in mmseg default setting, but now it may not used with latest tuned parameters. After checking it, let me revert it to original one. Thanks for your feedback.

mpa/cls/stage.py
supersoob: After reading your comment, I thought about the case that the supervised should be run. I'd say that when model classes and data classes are totally different without intersection, supervised learning needs to be initiated. Otherwise, should start incremental learning.
So I plan to change this part
`if len(set(model_classes) & set(dst_classes))==0: #change loss to cross entropy`
what do you think? have any better idea?
And I will change the default op into "REPLACE" to get the dst_classes correctly as you discussed in your previous PR.

harimkang: I've been thinking about this for a little bit..
For supervised learning, I think there are two cases:
- where model_classes and data_classes are the same
- or model_classes = []

```
if len(model_classes) == 0 or model_classes == data_classes:
  cfg.model.head.loss = dict(type='CrossEntropyLoss', loss_weight=1.0)
else:
  cfg.model.head.loss = ConfigDict(
      type='SoftmaxFocalLoss',
      loss_weight=1.0,
      gamma=gamma,
      reduction='none',
  )
```
In this case, I think you might have to look at the order of model_classes and data_classes (model_classes == data_classes).

There may be some parts that I missed, so please let me know if there are any.

supersoob: Agree that these two cases are for supervised learning.

- where model_classes and data_classes are the same -> No new classes thus not required incremental learning

- or model_classes = [] -> Model never been trained

And in my opinion, I think the case I mentioned in the last comment should be considered, too.

- len(set(model_classes) & set(dst_classes))==0 -> for example, the model has trained with some or large-scale data, but want to optimize it to my custom data

For example, there is a model trained with label [a,b], but the label schema from data is [c,d].
And they are totally different and no intersection here.
In that case, I'd say this is not incremental and the model is to run in the way of supervised learning/finetune.
Plus, since we already consented to enable REPLACE mode, the dst_classes will be replaced by [c,d] and the newly trained model's label schema will be turned to [c,d]. If this case is deemed as incremental, MERGE op should be set.

I don't know the correct answer about dealing the learning type, since there are so many cases and perspectives.
If you have any other opinion, please comment freely:)

harimkang: For our current use case, we can consider whether the label_schema changes or not.
- model_classes = [] -> SL
- model_classes == data_classes -> SL
- model_classes is not empty and label_schema changes -> C-IL
- but, [a, b] -> [c, d] case means you need to learn like SL, right?

Then, I think it's a good idea to check also whether it's MERGE or REPLACE. What do you think? (However, I think we need to find the case that is both REPLACE and Incremental.)


we need to write code to properly capture these all cases.

supersoob: - model_classes = [] -> SL from scratch **-> Yes**
- model_classes is not empty and label_schema changes -> C-IL **-> Yes, but may have a problematic case like you mentioned (REPLACE + Incremental)**
- but, [a, b] -> [c, d] case means you need to learn like SL, right? **-> Yes**

To clearly explain the third case,

-  model class = [a,b]
-  data class = [c,d]

Want to deal as incremental?
dst_classes = [a,b,c,d] (MERGE op required in current code base)

Want to deal as supervised?
dst_classes = [c,d] (REPLACE op required in current code base)

I think this should be deemed supervised learning case. Do you agree with it?

However, I found a problematic case in both REPLACE and Incremental mode you mentioned.
When model classes and data classes have same classes partially, like
model class = [a,b]
data class = [b,c]
dst_classes = [b,c] (REPLACE op)
dst_classes = [a,b,c] (MERGE op)

final dst_classes will be same as data classes when REPLACE op is enabled. If it learns in incremental mode, it would not learn some original model classes, which are not included in data classes (and actually it is not in data folder). In SC, since the data folder path is constant, we don't need to worry if data class is not a superset of model_class (or maybe not when classes are removed from the user). But when pre-trained model is loaded manually and partially has same classes with dataset, I wonder which learning mode the user wants (run incrementally? or from scratch with dataset?). If the case is for supervised learning, we can leave REPLACE op as default and need additional 'if condition' to deal with this loss change code. If not, MERGE op is needed only for this case. Which op need to choose? What do you think?

I hope @goodsong81 @JihwanEom @chuneuny-emily @jaegukhyun join this discussion and suggest ideas if you have time. If I missed something or understand in wrong way, please let me know or correct me.

harimkang: I think the most appropriate mode right now is 'REPLACE'. The selection of 'from scratch' can be made through the check box in the SC UI, so I don't think it's a consideration.

Rather, I think we can think about whether to take the 3rd case with incremental mode or supervision mode.

supersoob: Then disabling REPLACE and turning on MERGE is required for incremental learning in the problematic case I mentioned. I realized that it's orthogonal to the user's decision made by SC checkbox (whether scratch or not) in OTE code. Could you share your opinion about the third case or else?

mpa/modules/hooks/early_stopping_hook.py
chuneuny-emily: Removed unnecessary code changes.

mpa/cls/stage.py
supersoob: I changed the codes here reflecting this discussion. Thank you for giving meaningful comments:)

mpa/cls/stage.py
```diff
@@ -297,17 +294,15 @@ def refine_tasks(train_cfg, meta, adapt_type):
     return model_tasks, old_tasks
 
 
-def refine_cls(train_cfg, meta, adapt_type):
+def refine_cls(train_cfg, data_classes, meta, adapt_type):
```
supersoob, line 297: I changed refine_cls and some codes to enable REPLACE mode in a right way. Before I changed this, REPLACE mode did not work at all in incremental learning. I ask for your review :)

models/segmentation/_base_/ocr_litehrnet18_mod2.py
```diff
@@ -0,0 +1,91 @@
+__norm_cfg = dict(type='BN', requires_grad=True)
+model = dict(
+    type='ClassIncrSegmentor',
+    is_task_adapt=True,
+    num_stages=1,
```
JihwanEom: Can we use EncoderDecoder class if we use only FCN head?

models/segmentation/_base_/ocr_litehrnet18_mod2.py
```diff
@@ -0,0 +1,91 @@
...
+                      sampler=dict(type='MaxPoolingPixelSampler', ratio=0.25, p=1.7),
+                      loss_weight=1.0),
+             ]),
+    ],
+    train_cfg=dict(
+        mix_loss=dict(
+            enable=False,
+            weight=0.1
+        ),
+        loss_reweighting=dict(
```
JihwanEom: I'm not sure this loss_reweighting config is working but I believe this should be removed.

models/segmentation/_base_/ocr_litehrnet18_mod2.py
```diff
@@ -0,0 +1,91 @@
...
+             enable_out_norm=False,
+             loss_decode=[
+                 dict(type='CrossEntropyLoss',
+                      use_sigmoid=False,
+                      sampler=dict(type='MaxPoolingPixelSampler', ratio=0.25, p=1.7),
+                      loss_weight=1.0),
+             ]),
+    ],
+    train_cfg=dict(
+        mix_loss=dict(
```
JihwanEom: Could you check the operation of mix_loss? I presume we might not need this option because we are using only one loss.

models/segmentation/_base_/ocr_litehrnet_s_mod2.py
```diff
@@ -0,0 +1,95 @@
...
+             enable_aggregator=True,
+             aggregator_merge_norm=None,
+             aggregator_use_concat=False,
+             enable_out_norm=False,
+             enable_loss_equalizer=True,
+             loss_decode=[
+                 dict(type='CrossEntropyLoss',
+                      use_sigmoid=False,
+                      sampler=dict(type='MaxPoolingPixelSampler', ratio=0.25, p=1.7),
+                      loss_weight=4.0),
```
JihwanEom: This loss_weight is also suspicious.

models/segmentation/_base_/ocr_litehrnet_x_mod3.py
```diff
@@ -0,0 +1,97 @@
...
+            weight=0.1
+        ),
+        loss_reweighting=dict(
+            weights={'decode_0.loss_seg': 0.9},
+            momentum=0.1
+        ),
+    ),
+    test_cfg=dict(
+        mode='whole',
+        output_scale=10.0,
```
JihwanEom: All models (18-mod2, s-mod2, x-mod3) are using same output_scale in config, but I'm not sure this is right config for each architecture. 
Could you check whether output_scale config is working correctly?

mpa/modules/models/classifiers/sam_classifier.py
```diff
@@ -82,6 +83,7 @@ def forward_train(self, img, gt_label, **kwargs):
         if self.multilabel:
             loss = self.head.forward_train(x, gt_label, **kwargs)
         else:
+            gt_label = torch.squeeze(gt_label)
```
JihwanEom: How about `gt_label.squeeze()` to prevent importing additional dependency.

recipes/stages/_base_/logs/tensorboard_logger.py
```diff
@@ -4,7 +4,7 @@
 log_config = dict(
     interval=100,
     hooks=[
-        dict(type='TextLoggerHook'),
+        dict(type='TextLoggerHook', ignore_last=False),
```
JihwanEom, line 7: I'm thinking about more intuitive name for this.. how about `force_print=True`

recipes/stages/segmentation/train.py
```diff
@@ -12,7 +12,16 @@
     weight_decay=0.0001,
 )
 
-lr_config = dict(min_lr=1e-06)
+lr_config = dict(
+    policy='ReduceLROnPlateau',
+    metric='mIoU',
```
JihwanEom: Is the current segmentation task is looking 'mIoU' to return the best checkpoint? I'm confused with 'mDice' metric.

recipes/stages/_base_/logs/tensorboard_logger.py
supersoob: This is a built-in parameter from mmcv. I found that it has supported for logging last_iter.

recipes/stages/segmentation/train.py
chuneuny-emily: It should be changed to mDice. I changed it.

recipes/stages/segmentation/train.py
```diff
@@ -23,15 +32,15 @@
...
                  r'\w*[.]?decode_head\.', r'\w*[.]?auxiliary_head\.']
 )
 
 custom_hooks = [
     dict(type='LazyEarlyStoppingHook',
-         patience=5,
-         iteration_patience=1000,
+         patience=8,
+         iteration_patience=0,
          metric='mIoU',
```
JihwanEom: @chuneuny-emily then, I think this metric also should be changed with 'mDice', am I right?

chuneuny-emily: Yes. I changed it too.

recipes/stages/segmentation/class_incr.py
```diff
@@ -5,10 +5,11 @@
 ]
 
 optimizer = dict(
-    type='SGD',
+    _delete_=True,
+    type='Adam',
```
JihwanEom, line 9: Adam shows better results than SGD? I remember you said SGD is better than Adam.

models/segmentation/_base_/ocr_litehrnet_x_mod3.py
chuneuny-emily: That configs are the default config on mmseg. Is there any reason you think that is not right config for each architecture?

recipes/stages/segmentation/class_incr.py
chuneuny-emily: SGD gave slight accuracy gain w/ 120 image set, but it affected too slow training time and accuracy drop on large backbone w/ 12 image set. With the small set, Adam is giving very fast learning w/ slight accuracy drop.

models/segmentation/_base_/ocr_litehrnet_x_mod3.py
JihwanEom: When I tested the performance of segmentation task, I remember output_scale had huge effect.
I'm not sure this `output_scale=10.0` is right config without OCR head and also wondering about whether this option is working correctly.

models/segmentation/_base_/ocr_litehrnet_s_mod2.py
chuneuny-emily: 4.0 is default value on mmseg. Is there any reason you think it's suspicious?

recipes/stages/segmentation/class_incr.py
JihwanEom: Okay, thank you for detailed explanation. Could you confirm grad_norm for segmentation tasks?
I remember OTEseg team applied Adam optimizer to stabilize grad_norm of very early epoch.

models/segmentation/_base_/ocr_litehrnet_s_mod2.py
JihwanEom: Yes, I think this can cause unstable training by multiplying 4 to origin loss value. In OTEseg s_mod2 model, it is using 'CrossEntropyLoss', 'GeneralizedDiceLoss', 'AMSoftmaxLoss', each weight is 4.0, 4.0, 1.0 respectively. 
So, I'm suspicious `loss_weight=4.0` can be too large without OCR head/Aux head.

recipes/stages/segmentation/class_incr.py
chuneuny-emily: grad_norm works w/ Adam, and does not work w/ SGD.

models/segmentation/_base_/ocr_litehrnet_s_mod2.py
chuneuny-emily: Thanks for your feedback. Since 'GeneralizedDiceLoss', 'AMSoftmaxLoss' were removed, I changed loss_weight to 1. I did sanity check and no effect on accuracy.

models/segmentation/_base_/ocr_litehrnet18_mod2.py
chuneuny-emily: Thanks for your feedback. Since there is only CE loss, mix_loss is not required. I removed it.

models/segmentation/_base_/ocr_litehrnet18_mod2.py
chuneuny-emily: Thanks for your feedback. Since there is only CE loss, mix_loss is not required. I removed it.

models/segmentation/_base_/ocr_litehrnet_x_mod3.py
chuneuny-emily: It is working, if I remove it, on large, new class was not learned. About the value 10, I'll conduct the experiments whether it's proper default per model.

models/segmentation/_base_/ocr_litehrnet18_mod2.py
chuneuny-emily: I updated it.

mpa/modules/models/segmentors/class_incr_segmentor.py
```diff
@@ -45,9 +45,8 @@ def load_state_dict_pre_hook(model, model_classes, chkpt_classes, chkpt_dict, pr
 
         model_dict = model.state_dict()
         param_names = [
-            'decode_head.0.conv_seg.weight',
-            'decode_head.0.conv_seg.bias',
-            'decode_head.1.conv_seg.weight',
+            'decode_head.conv_seg.weight',
+            'decode_head.conv_seg.bias',
```
goodsong81, line 49: Is this enough for arch change?
Can we load old OTE model weights from, for example, SC1.1 model with OCR head?

goodsong81 (approved):

mpa/modules/models/segmentors/class_incr_segmentor.py
chuneuny-emily: Currently it does not support. Let me update it on next PR.

PR #23, opened 2022-07-04
Update classification task adapt flow

JihwanEom:
### Summary
- Prevent [data_cls] -> [data_cls] training when run without previous label schema.


JihwanEom:
merged with https://github.com/openvinotoolkit/model_preparation_algorithm/pull/22

PR #24, opened 2022-07-06
[Detection] Output saliency map and feature vector for MPA Det

eugene123tw:


mpa/det/inferrer.py
```diff
@@ -32,6 +34,8 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
         self._init_logger()
         mode = kwargs.get('mode', 'train')
         eval = kwargs.get('eval', False)
+        dump_features = kwargs.get('dump_features')
+        dump_saliency_map = kwargs.get('dump_saliency_map')
```
goodsong81: Why don't you add the default value 'False' as above?

goodsong81 (changes requested):
Thank you for the PR.
I'd like to ask a minor change.

mpa/det/inferrer.py
eugene123tw: Added. Thanks!

goodsong81 (approved):
LGTM!

PR #25, opened 2022-07-07
Integrate SSD and YOLOX model

jaegukhyun:
- Add SSD and YOLOX model
- Change template_experimental.yaml to template.yaml
- Change optimization hyperparameters
- Change learning schedule hyperparameters


mpa/det/stage.py
```diff
@@ -113,6 +113,15 @@ def configure_data(self, cfg, training, **kwargs):
                         seed=cfg.seed
                     )
                 )
+            if 'dataset' in cfg.data.train:
+                train_cfg = self.get_train_data_cfg(cfg)
+                train_cfg.ote_dataset = cfg.data.train.ote_dataset
+                train_cfg.labels = cfg.data.train.labels
+                train_cfg.data_classes = cfg.data.train.data_classes
+                train_cfg.new_classes = cfg.data.train.new_classes
+                cfg.data.train.pop('ote_dataset')
```
harimkang: Is there a better way to do it than this? Wouldn't there also be a better location?
And this code seems to be error-prone without 'ote_dataset', 'data_classes', and 'new_classes' in train_data_cfg. 
It would be safe if you could declare the default value.

mpa/det/stage.py
```diff
@@ -251,7 +260,13 @@ def configure_task_cls_incr(self, cfg, task_adapt_type, org_model_classes, model
...
                 alpha = 0.75
                 gamma = 1 if cfg['task_adapt'].get('efficient_mode', False) else 2
-
+            elif cfg.model.bbox_head.type in ['YOLOXHead', 'CustomYOLOXHead']:
+                if cfg.data.train.type == 'MultiImageMixDataset':
+                    cfg.data.train.pop('ann_file', None)
+                    cfg.data.train.pop('img_prefix', None)
+                    cfg.data.train['labels'] = cfg.data.train.pop('labels', None)
+                    self.add_yolox_hooks(cfg)
+                cfg.ignore = False  # YOLOX does not yet support ignore mode
```
harimkang: I think SSD needs to be processed in the same way.

mpa/det/trainer.py
```diff
@@ -153,7 +155,12 @@ def train_worker(gpu, target_classes, datasets, cfg, distributed=False,
         # model
         model = build_detector(cfg.model)
         model.CLASSES = target_classes
-
+        # Do clustering for SSD model
+        if hasattr(cfg.model, 'bbox_head') and hasattr(cfg.model.bbox_head, 'anchor_generator'):
+            if getattr(cfg.model.bbox_head.anchor_generator, 'reclustering_anchors', False):
+                train_cfg = Stage.get_train_data_cfg(cfg)
+                train_dataset = train_cfg.ote_dataset
```
harimkang: Does this train_cfg always have the 'ote_dataset'?

mpa/modules/datasets/samplers/cls_incr_sampler.py
```diff
@@ -63,4 +64,4 @@ def __iter__(self):
         return iter(indices)
 
     def __len__(self):
-        return self.data_length * self.repeat
+        return self.data_length * self.repeat
```
harimkang: I think the checking for flake8 should be done for all files related to this PR.

mpa/modules/models/heads/__init__.py
```diff
@@ -3,3 +3,4 @@
 #
 
 # flake8: noqa
+from . import custom_yolox_head
```
harimkang: As far as I know, this has been moved to a different location without importing from here. 
I think this import statement should also be moved to another place.

recipes/stages/_base_/data/coco_ote_yolox.py
```diff
@@ -0,0 +1,45 @@
...
+]
+
+__dataset_type = 'CocoDataset'
+__data_root = 'data/coco/'
+
+__train_pipeline = {{_base_.train_pipeline}}
+__test_pipeline = {{_base_.test_pipeline}}
+__img_scale = {{_base_.img_scale}}
+
+__samples_per_gpu = 2
```
harimkang: Is there a reason why batch size is 2 by default?

recipes/stages/_base_/schedules/plateau.py
```diff
@@ -3,8 +3,8 @@
 lr_config = dict(
     policy='ReduceLROnPlateau',
     metric='bbox_mAP',
-    patience=3,
-    iteration_patience=600,
+    patience=30,
```
harimkang, line 6: I think it's too long, but it's the test result, right?

recipes/stages/detection/imbalance_ssd.py
```diff
@@ -0,0 +1,35 @@
+_base_ = [
```
harimkang: It seems that there is no need to take the recipe of ssd and yolox separately. 
On the OTE side, there is a way to put data_pipeline in the config, so I recommend you to change it to that way, 
and these seem to be better to erase.

harimkang (changes requested):

mpa/modules/models/detectors/custom_yolox_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+            # Mix weights
+            model_param = model_dict[model_name].clone()
+            chkpt_param = chkpt_dict[chkpt_name]
+            for m, c in enumerate(model2chkpt):
+                if c >= 0:
+                    model_param[m].copy_(chkpt_param[c])
+
+            # Replace checkpoint weight by mixed weights
+            chkpt_dict[chkpt_name] = model_param
```
JihwanEom: Please check flake8 convention.

mpa/modules/utils/task_adapt.py
```diff
@@ -139,3 +139,13 @@ def sync_transpose_tensor(src_w, dst_w):
         return src_w.t()
     else:
         raise ValueError('the size of model and checkpoint file are not matched')
+
+
+def get_dataset_without_wrapping(dataset):
```
JihwanEom: How about 'unwrap_dataset'

recipes/stages/_base_/data/coco_ote_yolox.py
```diff
@@ -0,0 +1,45 @@
...
+        img_prefix=__data_root + 'val2017/',
+        test_mode=True,
+        pipeline=__test_pipeline),
+    test=dict(
+        type=__dataset_type,
+        ann_file=__data_root + 'annotations/instances_val2017.json',
+        img_prefix=__data_root + 'val2017/',
+        test_mode=True,
+        pipeline=__test_pipeline)
+)
```
JihwanEom: Need to check flake8

recipes/stages/_base_/data/coco_ote_yolox.py
```diff
@@ -0,0 +1,45 @@
...
+__train_pipeline = {{_base_.train_pipeline}}
+__test_pipeline = {{_base_.test_pipeline}}
+__img_scale = {{_base_.img_scale}}
+
+__samples_per_gpu = 2
+
+data = dict(
+    samples_per_gpu=__samples_per_gpu,
+    workers_per_gpu=4,
+    num_classes=2,
```
JihwanEom: Are other templates also using 2 as the default value of num_classes?

recipes/stages/_base_/data/pipelines/coco_ote_yolox_pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+        flip=False,
+        transforms=[
+            dict(type='Resize', keep_ratio=True),
+            dict(type='RandomFlip'),
+            dict(type='Pad', size=(416, 416), pad_val=114.0),
+            dict(type='Normalize', **img_norm_cfg),
+            dict(type='DefaultFormatBundle'),
+            dict(type='Collect', keys=['img'])
+        ])
+]
```
JihwanEom: Please check Flake8

recipes/stages/_base_/data/pipelines/coco_ote_yolox_pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+test_pipeline = [
+    dict(type='LoadImageFromFile'),
+    dict(
+        type='MultiScaleFlipAug',
+        img_scale=(416, 416),
+        flip=False,
+        transforms=[
+            dict(type='Resize', keep_ratio=True),
+            dict(type='RandomFlip'),
+            dict(type='Pad', size=(416, 416), pad_val=114.0),
```
JihwanEom: Why use 114.0 as pad_val? Is it common sense for detection tasks? It's just a question.

recipes/stages/_base_/models/detectors/yolox.custom.py
```diff
@@ -0,0 +1,10 @@
+_base_ = [
+    './yolox.py'
+]
+
+model = dict(
+    type='CustomYOLOX',
+    bbox_head=dict(
+        type='CustomYOLOXHead',
+    ),
+)
```
JihwanEom, line 10: No new line

recipes/stages/_base_/models/detectors/yolox.py
```diff
@@ -0,0 +1,31 @@
...
+            type='SimOTAAssigner',
+            center_radius=2.5)),
+    # In order to align the source code, the threshold of the val phase is
+    # 0.01, and the threshold of the test phase is 0.001.
+    test_cfg=dict(
+        score_thr=0.01,
+        nms=dict(type='nms', iou_threshold=0.65),
+        max_per_img=100)
+)
+load_from = 'https://storage.openvinotoolkit.org/repositories/openvino_training_extensions/models/object_detection/v2/yolox_tiny_8x8.pth'
```
JihwanEom, line 31: No new line

samples/cfgs/models/detectors/yolox_darknet.yaml
```diff
@@ -0,0 +1,3 @@
+_base_: [
+  '../../../../recipes/stages/_base_/models/detectors/yolox.custom.py'
+]
```
JihwanEom, line 3: No new line

mpa/det/trainer.py
```diff
@@ -153,7 +155,12 @@ def train_worker(gpu, target_classes, datasets, cfg, distributed=False,
         # model
         model = build_detector(cfg.model)
         model.CLASSES = target_classes
-
+        # Do clustering for SSD model
+        if hasattr(cfg.model, 'bbox_head') and hasattr(cfg.model.bbox_head, 'anchor_generator'):
```
JihwanEom, line 159: How about `cfg.model.get('bbox_head', False)` following the convention of other tasks?

mpa/modules/utils/task_adapt.py
harimkang: I think the current function name looks better. What do you think?

mpa/det/trainer.py
jaegukhyun: I just followed mmdetection's rule.

mpa/det/trainer.py
jaegukhyun: As far as I know, yes

mpa/modules/utils/task_adapt.py
jaegukhyun: I think unwrap_dataset is more intuitive

recipes/stages/_base_/data/coco_ote_yolox.py
jaegukhyun: I think that we should discuss about this a little. 
In configure_hyperparams function(mpa/det/stage.py) line 334  "cfg.data.samples_per_gpu = bs" define samples_per_gpu as batch size. I think samples_per_gpu should be bs/gpu_num and __samples_per_gpu don't need to be  defined by config. What do you think?

recipes/stages/_base_/data/coco_ote_yolox.py
jaegukhyun: By coco_ote.py, yes. But as I mentioned above we should think about definition of __samples_per_gpu

recipes/stages/_base_/data/pipelines/coco_ote_yolox_pipeline.py
jaegukhyun: I just followed mmdetection implementation. But It should be consider defining proper pad size for Mosaic augmentation. As far as I know, mosaic augmentation is very common to YOLO style models. I'll check pad value
from original yolo implementation

recipes/stages/_base_/schedules/plateau.py
jaegukhyun: It is test result, but for shortening schedule it will be changed

mpa/det/stage.py
jaegukhyun: from mpa/det/stage.py line#248
            if bbox_head.type in ['SSDHead', 'CustomSSDHead']:
                gamma = 1 if cfg['task_adapt'].get('efficient_mode', False) else 2
                bbox_head.type = 'CustomSSDHead'
                bbox_head.loss_cls = ConfigDict(
                    type='FocalLoss',
                    loss_weight=1.0,
                    gamma=gamma,
                    reduction='none',
                )
I think it is already implemented. @harimkang What do you think

mpa/det/stage.py
jaegukhyun: I think data_classes and new_classes are always defined in task.py(mpa_tasks/apis)
Also ote_dataset is always defined in task.py(mpa_tasks/apis/detection/)
Therefore train.data.cfg always have these attributes.
The location could be arguable. I followed agreement from group tech review.
@harimkang What do you think?

mpa/det/trainer.py
harimkang: i think `train_dataset = train_cfg.get('ote_dataset', None)` is better.

mpa/det/stage.py
harimkang: recipes/stages/_base_/models/detectors/ssd.custom.py -> ignore = False
Isn't the above variable sufficient? (SSD, YOLOX) 
Is there a reason to apply it separately in the code?

mpa/det/trainer.py
harimkang: I think the original code looks better. (I think it doesn't need to be modified.)

mpa/det/trainer.py
JihwanEom: Did you test export cli? train_cfg can be None if cfg is empty.

mpa/det/stage.py
jaegukhyun: No, my mistakes

mpa/det/trainer.py
jaegukhyun: Do export need trainer? I think export procedure do not come through this code

harimkang (approved):
LGTM

goodsong81 (approved):
LGTM.

PR #26, opened 2022-07-07
[Segmentation] Output feature vector for MPA seg algo

eugene123tw:


mpa/seg/inferrer.py
```diff
@@ -108,16 +109,26 @@ def infer(self, cfg):
...
+        SegStage.set_inference_progress_callback(eval_model, cfg)
+
+        segmentations = []
+        feature_vectors = []
+        for data in mm_val_dataloader:
+            with torch.no_grad():
+                result, feature_vector = eval_model(return_loss=False, output_logits=True, **data)
+                segmentations.append(result)
+                feature_vectors.append(feature_vector)
+            assert len(result) == 1
```
goodsong81: - Let's avoid inserting assert in product code; explicitly raise exception.
- Simple loop looks concise, but it will lose batch_size>1 execution. single_gpu_test() seems support it.
- In addition, if you perform grep 'return_repr_vector' you would find some matches; could you check if the 'feature vector' functionality is already implemented in ote mmseg?

goodsong81 (commented):
I'm OK with this change. Left some comments.
@JihwanEom Could you look into this PR?

mpa/seg/inferrer.py
```diff
@@ -108,16 +109,26 @@ def infer(self, cfg):
             _ = load_checkpoint(model, cfg.load_from, map_location='cpu')
 
         # Inference
-        model = MMDataParallel(model, device_ids=[0])
+        model.eval()
+        eval_model = MMDataParallel(model, device_ids=[0])
```
JihwanEom: Is CPU inference considered? I think device_ids argument can cause some errors with cpu-only machine.

JihwanEom (approved):
LGTM

chuneuny-emily (approved):

PR #27, opened 2022-07-08
Enable hierarchical classification in MPA

JihwanEom:
### Summary
- Add hierarchical model templates for MPA
- Add CustomHierarchicalLinearClsHead/NonLinearClsHead
- Updated stage.py to support hierarchical classification
- Add data_loader config capacity to set drop_last option


mpa/cls/trainer.py
```diff
@@ -158,7 +160,8 @@ def train_worker(gpu, dataset, cfg, distributed, validate, timestamp, meta):
                         num_gpus=len(cfg.gpu_ids),
                         dist=distributed,
                         round_up=True,
-                        seed=cfg.seed
+                        seed=cfg.seed,
+                        drop_last=train_data_cfg.drop_last if train_data_cfg.get('drop_last', False) else False
```
harimkang: Is there a reason to add this?

harimkang (commented):
LGTM

mpa/cls/trainer.py
JihwanEom: Yes, default batch size for classification task is 32 and the minimum sample number for SC classification is only 18.
In this case, if we set drop_last=True, the image never comes into forward method. (only dry-run). I handled this case in task.py (https://github.com/openvinotoolkit/training_extensions/blob/jeom/enable-mpa-hierarchical/external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py#L280)

But with hierarchical classification, this task requires drop_last=True, for stable hierarchical information indexing.
So, I add this capacity to avoid the situation I mentioned above.

mpa/modules/models/heads/custom_hierarchical_linear_cls_head.py
```diff
@@ -0,0 +1,138 @@
...
+import torch.nn as nn
+from mmcv.cnn import normal_init
+
+from mmcls.models.builder import HEADS, build_loss
+from mmcls.models.heads import MultiLabelClsHead
+
+
+@HEADS.register_module()
+class CustomHierarchicalLinearClsHead(MultiLabelClsHead):
+    """Custom Linear classification head for multilabel task.
```
jaegukhyun: Isn't it head for "hierarchical task"? not "multilabel task"

JihwanEom: I'll revise it, thank you.

mpa/modules/models/heads/custom_hierarchical_non_linear_cls_head.py
```diff
@@ -0,0 +1,162 @@
...
+import torch.nn as nn
+from mmcv.cnn import normal_init, constant_init, build_activation_layer
+
+from mmcls.models.builder import HEADS, build_loss
+from mmcls.models.heads import MultiLabelClsHead
+
+
+@HEADS.register_module()
+class CustomHierarchicalNonLinearClsHead(MultiLabelClsHead):
+    """Custom Linear classification head for multilabel task.
```
jaegukhyun: I think description for this function is not match
Also, NonLinearClsHead is not intuitive. How about MLPHead?

JihwanEom: I don't agree with MLPHead, The difference between LinearClsHead and NonLinearClsHead cannot be summarized by MLPHead. But I'll think about more intuitive name. Thank you for your suggestion.

mpa/modules/models/classifiers/sam_classifier.py
```diff
@@ -32,6 +32,8 @@ class SAMImageClassifier(ImageClassifier):
     def __init__(self, task_adapt=None, **kwargs):
         if 'multilabel' in kwargs:
             self.multilabel = kwargs.pop('multilabel')
+        if 'hierarchical' in kwargs:
```
harimkang, line 35: Is there a reason why you implemented this on the SAM Classifier? It looks better to use a new classifier that inherits the SAM Classifier.

mpa/cls/trainer.py
```diff
@@ -172,7 +175,8 @@ def train_worker(gpu, dataset, cfg, distributed, validate, timestamp, meta):
...
```
harimkang: I think it would be better to subtract the 'drop_last' variable so that it is declared separately.

harimkang (changes requested):
I left a few comments.

mpa/modules/models/classifiers/sam_classifier.py
JihwanEom: I agree with you. Can I put this work in backlog? I think that work should consider not only hierarchical, but also multilabel.

mpa/cls/trainer.py
JihwanEom: Do you mean like below:
`drop_last = train_data_cfg.drop_last if train_data_cfg.get('drop_last'. False) else False`
...
`build_dataloader(num_gpus, dist, ..., drop_last=drop_last)`?

harimkang (approved):
LGTM,
However, it seems necessary to remember that SAMClassifier is playing too many roles that do not fit the name, and to separate the roles later. Please create a task in the backlog.
(I think multi-label and h-label should have classifier respectively.)

goodsong81 (approved):
LGTM

PR #28, opened 2022-07-11
Enable mixed precision in cls task & Update mmcls version to v0.15.0

supersoob:
This PR includes

- Update mmcls version to v0.15.0 to enable FP16
- SAM optimizer for fp16


mpa/modules/hooks/__init__.py
```diff
@@ -20,3 +20,4 @@
 from . import logger_replace_hook
 from . import auxiliary_hooks
 from . import save_initial_weight_hook
+from . import fp16_sam_optimizer_hook
```
JihwanEom: No new line for this code

mpa/modules/hooks/fp16_sam_optimizer_hook.py
```diff
@@ -0,0 +1,99 @@
...
+                        continue
+                    e_w = param.grad * scale.to(param)
+                    param.add_(e_w)  # Climb toward gradient (increasing loss)
+                    param2move[param] = e_w  # Saving for param restore
+
+        # SAM gradient
+        runner.optimizer.zero_grad()
+        max_outputs = runner.model.train_step(current_batch, runner.optimizer)  # forward() on maxima param
+        max_loss = max_outputs['loss']
+        self.loss_scaler.scale(max_loss).backward()
```
JihwanEom, line 63: Could you explain why only SAM uses `self.loss_scaler` for backward? SGD gradient seems like to be calculated without `self.loss_scaler`

mpa/modules/hooks/__init__.py
supersoob: I can't understand, what do you mean by submodule?

mpa/cls/trainer.py
```diff
@@ -14,19 +14,20 @@
...
 from mmcv.parallel import MMDataParallel, MMDistributedDataParallel
-from mmcv.runner import DistSamplerSeedHook, build_optimizer, build_runner, HOOKS
+from mmcv.runner import DistSamplerSeedHook, Fp16OptimizerHook, build_optimizer, build_runner, HOOKS
 
 from mmcls import __version__
 from mmcls.datasets import build_dataset, build_dataloader
 from mmcls.models import build_classifier
 from mmcls.utils import collect_env
-from mmcls.core import (DistOptimizerHook, Fp16OptimizerHook)
+from mmcls.core import DistOptimizerHook
```
JihwanEom, line 23: `DistOptimizerHook` is not deprecated? It's just a question, but I think that should be considered on this PR.

JihwanEom (commented):
I left a few comments.

mpa/modules/hooks/__init__.py
JihwanEom: Sorry, I confused. Please add new line for this.

mpa/modules/hooks/fp16_sam_optimizer_hook.py
supersoob: Are you talking about SGD gradient in OptimizerHook? loss_scaler is needed for mixed precision. It prevents gradient from becoming zero in fp16 by multiplying by scale factor to loss. And weight gradients should be unscaled before updating optimizer.

JihwanEom: Yes, I agree but I'm wondering `curr_loss.backward()` is right code for mixed precision. Even though I'm not sure for all details about FP16, I think that may be `self.loss_scaler.scale(curr_loss).backward()`

supersoob: I was concerned about that, too. If we want to modify the parameter before the weight update, it should be unscaled to maintain the magnitude. But unscale_ must be called once per optimizer step. But the first backward in the sam optimizer is not for parameter update, so unscale function will be duplicated causing RuntimeError. The first backward in SAM is for moving parameter to local maximum and getting larger loss for the next bp, not performing the actual parameter updating. So I thought that zero gradient would not be matter to the first backward because it will be dealt by scaling in the next backward if it happens. If you have an other opinion, let me know.

mpa/cls/trainer.py
supersoob: We still using this. Please look at 241 line.

JihwanEom: I intended whether latest or 0.15.0 official mmclassification uses this hook or not removed yet. (I remember they removed deprecated things on 0.23.0)

harimkang (changes requested):
Has the OTE PR related to this PR been created?
This is a library version upgrade, so testing is essential along with the 'init_venv' modification. MPA PR should also be merged after passing the OTE test.

supersoob:
> Has the OTE PR related to this PR been created? This is a library version upgrade, so testing is essential along with the 'init_venv' modification. MPA PR should also be merged after passing the OTE test.

No change in OTE code. And I already have tested in OTE CI but didn't make OTE PR. I will make it after this is merged to develop.

mpa/cls/trainer.py
supersoob: I checked that it is still using in 0.15.0v.

mpa/modules/hooks/fp16_sam_optimizer_hook.py
```diff
@@ -0,0 +1,99 @@
...
+class Fp16SAMOptimizerHook(Fp16OptimizerHook):
+    '''Sharpness-aware Minimization optimizer hook
+
+    Implemented as OptimizerHook for MMCV Runners
+    - Paper ref: https://arxiv.org/abs/2010.01412
+    - code ref: https://github.com/davda54/sam
+    '''
+
+    def __init__(self, rho=0.05, start_epoch=1, **kwargs):
+        print(f"rho {rho}, st_epoch {start_epoch}, {kwargs}")
```
goodsong81: Please remove print() and use log if you think it's needed.

goodsong81 (changes requested):
Please refer to my comment. Others looks good to me.

goodsong81 (approved):
LGTM.

JihwanEom (approved):
LGTM

harimkang (approved):

PR #29, opened 2022-07-13
Enable Cross Focal Loss for CPU & Remove FP16 Inference

harimkang:
### Summary
- Resolve issues that stop at CPU when using cross sigmoid focal loss
- Remove FP16 wrapping in Inferrer.py for inference


goodsong81 (approved):
LGTM.

PR #32, opened 2022-07-15
MPA modification for HPO

eunwoosh:
This PR implemented as bellow

- save current_iters in after_epoch() and after_iter() of the OTEProgressHook class
- pass the curren_iters to time_monitor, then it's used by HPO.

### DIFF
--- a/mpa/modules/hooks/__init__.py
+++ b/mpa/modules/hooks/__init__.py
@@ -19,3 +19,4 @@ from . import early_stopping_hook
 from . import progress_update_hook
 from . import logger_replace_hook
 from . import auxiliary_hooks
+from . import save_initial_weight_hook
--- a/mpa/modules/hooks/early_stopping_hook.py
+++ b/mpa/modules/hooks/early_stopping_hook.py
@@ -200,6 +200,7 @@ class OTEProgressHook(Hook):
 
     @check_input_parameters_type()
     def after_epoch(self, runner: BaseRunner):
+        runner.log_buffer.output['current_iters'] = runner.iter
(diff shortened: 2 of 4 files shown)


mpa/modules/hooks/__init__.py
```diff
@@ -18,3 +18,4 @@
 from . import early_stopping_hook
 from . import progress_update_hook
 from . import logger_replace_hook
+from . import save_initial_weight_hook
```
harimkang: There seems to be a flake8 issue.

mpa/cls/trainer.py
```diff
@@ -113,18 +113,22 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
         # cfg.dump(osp.join(cfg.work_dir, 'config.yaml')) # FIXME bug to save
         # logger.info(f'Config:\n{cfg.pretty_text}')
 
+        # additional callback
+        save_model_callback = kwargs.get('save_model_callback')
```
harimkang: This is a callback for HPO with OTE and it seems not to be used in MPA, is it correct?

harimkang (changes requested):
I will approve it after solving flake8 issue :)

mpa/cls/trainer.py
eunwoosh: yes, you're right. I removed all of save_model_callback related things.

mpa/modules/hooks/__init__.py
eunwoosh: You mean f-srting issue in save_initial_weight_hook, right?

harimkang (approved):
LGTM

PR #33, opened 2022-07-15
[Detection] Change final validation ckpt as the best ckpt

jaegukhyun:
Change the final validation ckpt as the best epoch's weights

### DIFF
--- a/mpa/det/trainer.py
+++ b/mpa/det/trainer.py
@@ -135,10 +135,7 @@ class DetectionTrainer(DetectionStage):
 
         # Save outputs
         output_ckpt_path = osp.join(cfg.work_dir, 'latest.pth')
-        best_ckpt_path = glob.glob(osp.join(cfg.work_dir, 'best_segm_mAP*.pth'))
-        if len(best_ckpt_path) > 0:
-            output_ckpt_path = best_ckpt_path[0]
-        best_ckpt_path = glob.glob(osp.join(cfg.work_dir, 'best_mAP*.pth'))
+        best_ckpt_path = glob.glob(osp.join(cfg.work_dir, 'best_*.pth'))
         if len(best_ckpt_path) > 0:
             output_ckpt_path = best_ckpt_path[0]
         return dict(final_ckpt=output_ckpt_path)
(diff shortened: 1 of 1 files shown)


mpa/det/trainer.py
```diff
@@ -138,7 +138,7 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
         best_ckpt_path = glob.glob(osp.join(cfg.work_dir, 'best_segm_mAP*.pth'))
         if len(best_ckpt_path) > 0:
             output_ckpt_path = best_ckpt_path[0]
-        best_ckpt_path = glob.glob(osp.join(cfg.work_dir, 'best_bbox_mAP*.pth'))
+        best_ckpt_path = glob.glob(osp.join(cfg.work_dir, 'best_mAP*.pth'))
```
goodsong81: What do you think of using best_*.pth? Then we don't need to care about specific metrics.
@harimkang Would this idea work?

harimkang: As it is, detection and instance-segmentation use 'mAP' in common, so it is right to bring the weights, 
but I think it is necessary to modify it considering the change of the metric that goes into the save_best of evaluation.
(not 'bbox_mAP' or 'mAP' -> use save_best metric)
@jaegukhyun How is it?

harimkang: Would it be too complicated to designate the part where the model is stored as 'best.pth' in the evaluation?

jaegukhyun: So, @harimkang do you mean that best*.pth is proper for general situations?

harimkang: I meant that it should be 'best.pth' or 'f'best_{metric}*.pth', not best * .pth.

goodsong81: The actual file name for the best model is decided by mmdet not mpa; * is not only for metric but also epoch number. ex) It is being saved like best_mAP_epoch12.pth.

jaegukhyun: Than the latter option(f'best_{metric}*.pth') is only available for better generalization. @goodsong81, What do you think?
I think both best_* and f'best_{metric}' could work

harimkang (approved):
It seems that there is no problem with its behavior, but it would be better to modify it to specify more specifically to find the correct pth, 
such as f'best_{self.key_indicator}_{current}.pth' of external/mmdetection/submodule/mmdet/core/valuation/valuation/valuation/hooks.py::291

PR #36, opened 2022-07-25
Detection SL default sampler set RandomSampler

harimkang:
### This PR includes :
- Update default sampler in SL to RandomSampler


jaegukhyun (approved):
LGTM

PR #40, opened 2022-08-08
Tiling Module

eugene123tw:


mpa/cls/stage.py
```diff
@@ -81,10 +81,6 @@ def configure(self, model_cfg, model_ckpt, data_cfg, training=True, **kwargs):
             if cfg.model.get('multilabel', False) or cfg.model.get('hierarchical', False):
                 cfg.model.head.pop('topk', None)
 
-        # Other hyper-parameters
-        if cfg.get('hyperparams', False):
-            self.configure_hyperparams(cfg, training, **kwargs)
```
harimkang, line 89: Is there a reason to remove this? Also, configure_hyperparams still appears to exist in this file.

eugene123tw: I remove it because `configure_hyperparams` is not used. The actual "configure" occurs here as `_init_recipe_hparam`: https://github.com/openvinotoolkit/training_extensions/blob/develop/external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py#L183

harimkang (approved):

JihwanEom (approved):

PR #41, opened 2022-08-09
[CVS-83924][Detection] YOLOX detection training does not work with CPU

jaegukhyun:
This PR includes giving device parameter to a YOLOX custom hook called SyncRandomSizeHook
From this PR, the yolox model could be run by cpu

### DIFF
--- a/mpa/det/stage.py
+++ b/mpa/det/stage.py
@@ -3,6 +3,7 @@
 #
 
 import numpy as np
+import torch
 
 from mmcv import ConfigDict
 from mmdet.datasets import build_dataset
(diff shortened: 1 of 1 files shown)


harimkang (commented):
it works in SC?

jaegukhyun:
Not really, Should I build SC through cpu machine, and check the PR?

harimkang (approved):
It doesn't seem to have any other influence, so I think we can merge it right away. 
However, it seems necessary to check if it is actually fixed in SC.

goodsong81 (approved):

PR #43, opened 2022-08-11
Enable mixed precision for all tasks and model templates

supersoob:
This PR includes
- Fix export errors by setting exported model to fp32 precision mode
- Get fp16 arg to set new optimizer config in segmentation

### DIFF
--- a/mpa/seg/exporter.py
+++ b/mpa/seg/exporter.py
@@ -45,7 +45,10 @@ class SegExporter(SegStage):
                 model = model.cuda(cfg.gpu_ids[0])
             else:
                 model = model.cpu()
-            export_model(model, cfg, output_path, target='openvino', output_logits=True, input_format='bgr')
+            precision = kwargs.pop('precision', 'FP32')
+            print(precision)
+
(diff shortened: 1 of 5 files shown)


mpa/seg/exporter.py
```diff
@@ -45,7 +45,10 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
                 model = model.cuda(cfg.gpu_ids[0])
             else:
                 model = model.cpu()
-            export_model(model, cfg, output_path, target='openvino', output_logits=True, input_format='bgr')
+            precision = kwargs.pop('precision', 'FP32')
+            logger.info(f'Model will be exported with precision {precision}')
+
+            export_model(model, cfg, output_path, target='openvino', output_logits=True, input_format='bgr', precision=precision)
```
goodsong81, line 51: We may be careful for BGR / RGB.
I see that only seg is using this specific parameter.

goodsong81 (approved):
LGTM.
Just refer to my short comment.

JihwanEom (approved):
LGTM

PR #54, opened 2022-08-31
[CVS-90509] Resolve segmentation cpu training issue

JihwanEom:
### Summary:
- Resolve [CVS-90509] Semantic segmentation training with all templates doesn't work in cpu.

### DIFF
--- a/mpa/cls/trainer.py
+++ b/mpa/cls/trainer.py
@@ -29,6 +29,7 @@ from mpa.cls.stage import ClsStage
 from mpa.modules.hooks.eval_hook import CustomEvalHook, DistCustomEvalHook
 from mpa.modules.hooks.fp16_sam_optimizer_hook import Fp16SAMOptimizerHook
 from mpa.utils.logger import get_logger
+from ..utils.data_cpu import MMDataCPU
 
 logger = get_logger()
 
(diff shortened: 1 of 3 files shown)


chuneuny-emily (approved):

mpa/seg/train.py
```diff
@@ -139,3 +142,69 @@ def train_segmentor(model,
     elif cfg.load_from:
         runner.load_checkpoint(cfg.load_from)
     runner.run(data_loaders, cfg.workflow)
+
+
+# Temporary copy from mmdet
+# TODO: refactor
+def scatter_cpu(inputs):
```
harimkang: This appears to be the same as the one at train.py in classification. If it's the same code, it seems appropriate to transfer it to the utils folder.

harimkang (changes requested):

mpa/cls/trainer.py
```diff
@@ -29,6 +29,7 @@
 from mpa.modules.hooks.eval_hook import CustomEvalHook, DistCustomEvalHook
 from mpa.modules.hooks.fp16_sam_optimizer_hook import Fp16SAMOptimizerHook
 from mpa.utils.logger import get_logger
+from ..utils.data_cpu import MMDataCPU
```
harimkang: I think `from mpa.utils.data_cpu import MMDataCPU ` is better

chuneuny-emily (approved):

harimkang (approved):

PR #55, opened 2022-09-05
Add Equalized Focal Loss

harimkang:
### Summary:
- Add Equalized Focal Loss for Class-Imbalance training (Detection/Instance-segmentation)


harimkang:
This will be held at a later time when the PoC is running again.

PR #57, opened 2022-09-08
Add ignore_last flag on semgnetation text logger

JihwanEom:
Prevent possible bug : Training graph will not be visualized when number of training sample is smaller than batch size.


harimkang:
What is the background to this?  Thanks for updating the description.

jaegukhyun (approved):
LGTM

PR #60, opened 2022-09-14
Fix CVS-91469 sseg compatibility issue

chuneuny-emily:
### Summary:
- Fix CVS-91469 sseg compatibility issue by adding litehrnet18, which is a deprecated model template.


chuneuny-emily:
I manually tested train/eval/optimize/export works fine.

JihwanEom (changes requested):
Should we add this model template? As far as I know, 18 has same architecture with 18-mod2.

PR #61, opened 2022-09-18
Added switching the EMA on based on number of images

kprokofi:


mpa/modules/hooks/model_ema_v2_hook.py
```diff
@@ -34,22 +34,26 @@ class ModelEmaV2Hook(Hook):
             to update ema parameters. Defaults to 5.
     """
 
-    def __init__(self, ema_decay=0.999, interval=1, start_epoch=1, **kwargs):
+    def __init__(self, ema_decay=0.9995, interval=1, start_epoch=1, switching_criteria=2000, **kwargs):
```
sovrasov: Please add the new parameter to the docstring. Also, the name `switching_criteria` sounds a bit unclear: 
something like `data_len_threshold` reveals the role of this arg.

mpa/cls/trainer.py
```diff
@@ -153,10 +151,13 @@ def train_worker(gpu, dataset, cfg, distributed, validate, timestamp, meta):
 
         # updated to adapt list of dataset for the 'train'
         data_loaders = []
+        max_num_images = 0
         for ds in dataset:
             if isinstance(ds, list):
-                sub_loaders = [
-                    build_dataloader(
+                for sub_ds in ds:
+                    if len(sub_ds) > max_num_images:
```
sovrasov: max_num_images = max(max_num_images, len(sub_ds))

kprokofi: Thanks, that much better

sovrasov (approved):

PR #64, opened 2022-09-23
Replace to IB loss for incremental learning in multi-class classification task

supersoob:
This PR includes

- Implementation of IB loss hook and IB loss
- Move all heads used for multi-class classification to `custom_cls_head.py` (+ delete existing head file)
- Rename 'LinearHead' -> 'CustomLinearHead', 'NonLinearHead' -> 'CustomNonLinearHead'
- Replace focal loss to IB loss for incremental learning in `stage.py` multi-class classification task


mpa/modules/models/heads/ib_loss_head.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import torch
+from mmcls.models.builder import HEADS
+from mmcls.models.heads import LinearClsHead
+from mpa.modules.models.heads.non_linear_cls_head import NonLinearClsHead
+
+
+@HEADS.register_module()
+class IBLossHead(LinearClsHead):
```
harimkang: I don't understand that the head is added for the new loss.
I don't think this is the right way to do it if it's just to get the current epoch (loss per epoch). What do you think?
Does this have any other intentions?

mpa/modules/hooks/sam_optimizer_hook.py
```diff
@@ -22,6 +23,12 @@ def __init__(self, rho=0.05, start_epoch=1, **kwargs):
         if rho < 0.0:
             raise ValueError('rho should be greater than 0 for SAM optimizer')
 
+    def before_train_epoch(self, runner):
```
harimkang: Is there any other reason why this should happen in optimizer hook? 
The code written on fp16 and SAM looks the same, but isn't it enough to make it into a separate hook? (create 1 hook and enable?)

harimkang (changes requested):
I left a few things in the comments that I didn't understand.

mpa/modules/models/heads/ib_loss_head.py
supersoob: Because IB loss requires to have features before the last layer, not only the final output.

harimkang: It seems that you can cover it with a hook (example -> hook for get feature before the last layer) without a separate head. The reason why I am concerned about adding a head is that it feels strange in structure to change the head for each loss.

supersoob: I don't get it. Is there a way to reach to the feature before the last layer by hook? I thought hook is only involved in the after/before training, not in the middle of training iteration. Do you suggest to make other custom EpochBasedRunner to reach to the last layer?

mpa/modules/hooks/sam_optimizer_hook.py
supersoob: You saying that it's better to make new hook that functions only insertion of cur_epoch. Is it right?

mpa/cls/stage.py
```diff
@@ -204,17 +204,20 @@ def configure_task(cfg, training, model_meta=None, **kwargs):
...
                 if not cfg.model.get('multilabel', False) and not cfg.model.get('hierarchical', False):
                     efficient_mode = cfg['task_adapt'].get('efficient_mode', True)
-                    gamma = 2 if efficient_mode else 3
                     sampler_type = 'balanced'
 
                     if len(set(model_classes) & set(dst_classes)) == 0 or set(model_classes) == set(dst_classes):
                         cfg.model.head.loss = dict(type='CrossEntropyLoss', loss_weight=1.0)
                     else:
+                        cls_num_list = [train_data_cfg.num_data[data_cls] for data_cls in dst_classes]
+                        if cfg.model.head.type == 'NonLinearClsHead':
```
sovrasov: This logic is extremely unclear, so I agree with Harim, if introducing a new head can be avoided, it's better to do that. If we can't avoid adding new heads, one can create a config where `NonLinearClsHead` is implicitly written.

mpa/cls/stage.py
```diff
@@ -204,17 +204,20 @@ def configure_task(cfg, training, model_meta=None, **kwargs):
...
+                            cfg.model.head.type = 'NonLinearIBLossHead'
+                        else:
+                            cfg.model.head.type = 'IBLossHead'
                         cfg.model.head.loss = ConfigDict(
-                            type='SoftmaxFocalLoss',
-                            loss_weight=1.0,
-                            gamma=gamma,
-                            reduction='none',
+                            type='IBLoss',
```
sovrasov, line 213: Hardcoded loss is also confusing: why not implicitly add this to `.yaml` configs? Now one can use softmax only, but after these changes only `IBLoss` will be available. What if someone wants to experiment with softmax and add it to a config? The changes in the config won't be effective.

recipes/stages/classification/class_incr.yaml
```diff
@@ -23,6 +23,3 @@ task_adapt:
     type: 'mpa'
     op: 'REPLACE'
 
-custom_hooks:
-- type: NoBiasDecayHook
```
sovrasov, line 27: Why NBD is removed? It doesn't work with IB loss?

mpa/modules/hooks/sam_optimizer_hook.py
harimkang: I didn't mean exactly what structure to implement. 
but, what I'm just saying is that the current structure doesn't seem to need to be divided into two hooks.

mpa/modules/models/heads/ib_loss_head.py
harimkang: In fact, I didn't really understand how to use loss. But what I want to say is that it is not a good logic to add a new head according to loss, I agree with Vlad's comment above.

mpa/modules/hooks/task_adapt_hook.py
```diff
@@ -40,6 +40,8 @@ def __init__(self,
 
         logger.info(f'Task Adaptation: {self.src_classes} => {self.dst_classes}')
         logger.info(f'- Efficient Mode: {self.efficient_mode}')
+        logger.info(f'- Sampler type: {self.sampler_type}')
+        logger.info(f'- Sampler flag : {self.sampler_flag}')
```
JihwanEom: Please align whitespace with others

mpa/modules/models/heads/ib_loss_head.py
JihwanEom: I'm not sure for detailed IB loss implementation, but if the thing that you want is extracting feature vector from last layer, registering additional forward hook before model update can be help for you. Of course, forward hook will require overhead, but clear code might have more merits for us.

mpa/cls/stage.py
supersoob: I have no choice but hardcode this part to apply the IB loss only in incremental setting. The current code is already hardcoded for enabling focal loss for incremental setting(where new classes are added, thus it checks the difference between label schema of model and dataset). I agree that this hardcode part is somewhat problematic for experiments, so should be dealt when we are in refactoring stage next sprint.

recipes/stages/classification/class_incr.yaml
supersoob: Yes, it was not compatible with IB loss.

mpa/modules/models/heads/ib_loss_head.py
supersoob: forward hook is a little inefficient because it's like doing forward passes twice, causing large overhead. Getting feature vector before the last layer for IB loss is inevitable. There are two choice for this implementation. (If curious about detailed implementation, refer to [the official code](https://github.com/pseulki/IB-Loss/blob/b29b8b005f614292e9ae8b43903a6f7df8ef09a6/models/resnet_cifar.py#L323) or original paper) 1) Making IB loss classifier, 2) Making IB loss head. Since we are using SAMClassifier, 1) need to inherit SAMClassifier, which is more hardcoded method. 2) is the better choice, I think. Reflecting comments of @harimkang and @sovrasov, I will try to implicitly implement the head for IB loss by renaming or wrapping it as custom head. If have any comment, please let me know.

jaegukhyun (commented):
LGTM

jaegukhyun (approved):

mpa/modules/models/heads/custom_cls_head.py
```diff
@@ -0,0 +1,158 @@
...
+import torch.nn.functional as F
+from mmcv.cnn import normal_init, constant_init, build_activation_layer
+
+from mmcls.models.builder import HEADS
+from mmcls.models.heads import LinearClsHead
+from mmcls.models.heads.cls_head import ClsHead
+
+
+@HEADS.register_module()
+class CustomNonLinearClsHead(ClsHead):
```
harimkang: Shouldn't this inherit from NonLinearClsHead, not ClsHead?

models/classification/_base_/densenet_121_omz.py
```diff
@@ -12,7 +12,7 @@
     neck=dict(
         type='GlobalAveragePooling'),
     head=dict(
-        type='LinearClsHead',
+        type='CustomLinearClsHead',
```
harimkang: I think it's better to only change the target models (mv3, effnet).

mpa/cls/stage.py
```diff
@@ -204,18 +204,20 @@ def configure_task(cfg, training, model_meta=None, **kwargs):
...
 
                     if len(set(model_classes) & set(dst_classes)) == 0 or set(model_classes) == set(dst_classes):
                         cfg.model.head.loss = dict(type='CrossEntropyLoss', loss_weight=1.0)
                     else:
                         cfg.model.head.loss = ConfigDict(
-                            type='SoftmaxFocalLoss',
-                            loss_weight=1.0,
-                            gamma=gamma,
-                            reduction='none',
+                            type='IBLoss',
```
harimkang, line 213: Wouldn't it be appropriate to change it in the model recipe rather than the stage?
Also, is IB loss only used for class-incremental?

mpa/modules/hooks/ib_loss_hook.py
```diff
@@ -0,0 +1,40 @@
...
+        self.cls_num_list = None
+        self.dst_classes = dst_classes
+
+    def before_train_epoch(self, runner):
+        # get loss from model
+        model_loss = self._get_model_loss(runner)
+
+        # pass the number of data per class and current epoch to IB loss
+        dataset = runner.data_loader.dataset
+        num_data = self._get_num_data(dataset)
```
harimkang: Is this a value that changes every epoch?

mpa/modules/models/heads/non_linear_cls_head.py
```diff
@@ -1,88 +0,0 @@
-# Copyright (C) 2022 Intel Corporation
```
harimkang: I think it's good to leave this file as is, what do you think?

samples/cfgs/models/heads/linear_cls.yaml
```diff
@@ -1,6 +1,6 @@
 model:
     head:
-        type: LinearClsHead
+        type: CustomLinearClsHead
```
harimkang: Wouldn't it be better to create a new 'custom_linear_cls.yaml' file? (The same goes for non_linear.)

harimkang (commented):
Add new comments

mpa/modules/models/heads/custom_cls_head.py
supersoob: Original one is from non_linear_cls_head.py
image

It inherit ClsHead.

models/classification/_base_/densenet_121_omz.py
supersoob: Jihwan says it would be problem in omz model. I will revert and remain [non_linear_cls_head.py](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/64/files/daa790b4b61074c5a4ee115cb1c3faf6d7a27e3f#diff-cea1efa8c06cc9c65e5f674a99dbfff87cd744951fb9666ea67507c8e78427a0)

mpa/modules/hooks/ib_loss_hook.py
supersoob: Maybe no. I just thought of possibility that dataset can be resized after epochs.

mpa/cls/stage.py
supersoob: Since IB loss is used for c-il, I put it in stage.py. Please refer to my comment of Vlad's review above.

harimkang: I think the problem is that loss-overriding doesn't happen now (because hard-declaration in stage.py), so shouldn't this be fixed?

mpa/modules/hooks/ib_loss_hook.py
harimkang: If so, isn't this an unnecessary operation every epoch?

mpa/modules/models/heads/custom_cls_head.py
harimkang: If you leave a non_linear_cls_head this should inherit from it.

mpa/cls/stage.py
supersoob: Why do you think loss-overriding is not happening? I checked IB loss works in incremental setting.

harimkang: What I was saying was that adding a loss in the recipe should reflect the actual training, but currently it doesn't. So, it seems to be possible to solve this and add IB loss to the recipe.

mpa/modules/hooks/ib_loss_hook.py
JihwanEom: Could you explain more about `dataset can be resized after epochs?` I can't imagine that case :(

mpa/modules/models/losses/ib_loss.py
```diff
@@ -0,0 +1,59 @@
...
+@LOSSES.register_module()
+class IBLoss(CrossEntropyLoss):
+    def __init__(self, num_classes, start=5, alpha=1000., **kwargs):
+        """IB Loss
+        https://arxiv.org/abs/2110.02444
+
+        Args:
+            num_classes (int): Number of classes in dataset
+            start (int): Epoch to start finetuning with IB loss
+            alpha (float): Alpha for IB loss
```
JihwanEom: Could you update to describe what alpha works?

mpa/modules/models/losses/ib_loss.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        super(IBLoss, self).__init__(loss_weight=1.0)
+        assert alpha > 0
```
JihwanEom: How about use `raise` instead of `assert`

mpa/modules/hooks/ib_loss_hook.py
supersoob: I think it too far from this implementation. I fixed the codes reflecting all your reviews. Please check again.

supersoob:
CLI test
image
API test
image

models/classification/_base_/resnet18.py
```diff
@@ -9,7 +9,7 @@
         style='pytorch'),
     neck=dict(type='GlobalAveragePooling'),
     head=dict(
-        type='LinearClsHead',
+        type='CustomLinearClsHead',
```
JihwanEom: Could you revert resnets? We are not using resnet for classification, but iseg can be affected by this change.

JihwanEom (approved):
LGTM

harimkang (approved):

PR #66, opened 2022-09-29
multilabel classification enhancement

kprokofi:
This PR connected with https://github.com/openvinotoolkit/training_extensions/pull/1288
* Added angular asymmetric loss
* Added separate config for multilabel classification (for simpler modification and tasks independence)
* Added seeds freezing
* Proposed simple EMA heuristic and necessary methods for EMA implemented (saving checkpoints and validation)
* Added 1cycle support


recipes/stages/classification/class_incr.yaml
```diff
@@ -16,13 +16,16 @@ optimizer:
...
+task_adapt:
     type: 'mpa'
     op: 'REPLACE'
 
 custom_hooks:
 - type: NoBiasDecayHook
   priority: 45
+
+seed: 5
+deterministic: True
```
supersoob: Can you explain why seed freezing is needed? (including deterministic true)

recipes/stages/_base_/data/multicls_data.py
```diff
@@ -0,0 +1,10 @@
+_base_ = [
```
supersoob: I can't find the usage of this file. Why is this file for?

recipes/stages/classification/class_incr.yaml
kprokofi: We need to freeze seeds to have a reproducible code. The results on small data vary a lot (+-10%), it is not appropriate. We can't compare and measure the performance of the model because of the big random error.

recipes/stages/_base_/data/multicls_data.py
kprokofi: Yes, this file was used for different augmentations setup. According to the last experiments there is no need to change augmentations. So, I will delete this file.

models/classification/ote_efficientnet_v2_s_multilabel.yaml
```diff
@@ -11,3 +11,4 @@ model:
             type: AsymmetricLossWithIgnore
             gamma_pos: 0.0
             gamma_neg: 0.0
+            reduction: mean
```
sovrasov: Did you check that this change doesn't actually affect anything?

kprokofi: following the code it shouldn't change anything
but for safety I reverted it back

models/classification/ote_efficientnet_b0_multilabel_al.yaml
```diff
@@ -0,0 +1,18 @@
+_base_: _base_/ote_efficientnet.py
+
+model:
+    type: SAMImageClassifier
+    task: classification
+    backbone:
+        version: b0
+    head:
+        type: CustomMultiLabelNonLinearClsHead
```
JihwanEom, line 5: How about inheriting `ote_efficientnet_b0_multilabel.yaml`? It's updating only head config.

mpa/cls/trainer.py
```diff
@@ -166,10 +170,11 @@ def train_worker(gpu, dataset, cfg, distributed, validate, timestamp, meta):
                         seed=cfg.seed,
                         drop_last=drop_last,
                         persistent_workers=False
-                    ) for sub_ds in ds
-                ]
+                    )
+                    sub_loaders.append(sub_dl)
```
JihwanEom: `sud_loaders` is not defined

mpa/cls/trainer.py
```diff
@@ -230,6 +235,9 @@ def train_worker(gpu, dataset, cfg, distributed, validate, timestamp, meta):
         # an ugly walkaround to make the .log and .log.json filenames the same
         runner.timestamp = f'{timestamp}'
 
+        # remember the max number of train images. It is needed for EMA hook heuristic
```
JihwanEom: We can get the number of images in hook from `len(runner.data_loader.dataset`. Could you merge this line in updated EMA Hook?

mpa/modules/hooks/eval_hook.py
```diff
@@ -19,12 +19,13 @@ class CustomEvalHook(Hook):
...
     """
 
-    def __init__(self, dataloader, interval=1, by_epoch=True, **eval_kwargs):
+    def __init__(self, dataloader, interval=1, by_epoch=True, ema_start_eval=10, **eval_kwargs):
         if not isinstance(dataloader, DataLoader):
             raise TypeError('dataloader must be a pytorch DataLoader, but got'
                             f' {type(dataloader)}')
         self.dataloader = dataloader
         self.interval = interval
+        self.ema_start_eval = ema_start_eval
```
JihwanEom: How about `ema_eval_start_epoch`

mpa/modules/hooks/eval_hook.py
```diff
@@ -55,18 +60,25 @@ def after_train_iter(self, runner):
...
 
-    def evaluate(self, runner, results):
+    def evaluate(self, runner, results, results_ema=None):
         eval_res = self.dataloader.dataset.evaluate(
             results, logger=runner.logger, **self.eval_kwargs)
+        if results_ema:
+            eval_res_ema = self.dataloader.dataset.evaluate(
+                results_ema, logger=runner.logger, **self.eval_kwargs)
+            if self.call_score(eval_res_ema) > self.call_score(eval_res):
+                eval_res = eval_res_ema
```
JihwanEom: If we evaluate twice, how about to parse the performance of each model?
ex) `top-1_accuracy: 87.50, EMA_top-1_accuracy: 91.27, airplane_acc: 70.20, EMA_airplane_acc: 75.20', ...`

mpa/modules/hooks/checkpoint_hook.py
```diff
@@ -51,13 +51,17 @@ def __init__(self,
...
             if runner.save_ckpt:
+                if runner.save_ema_model:
+                    backup_model = runner.model
+                    runner.model = runner.ema_model
                 runner.logger.info(f'Saving checkpoint at {runner.epoch + 1} epochs')
                 if self.sync_buffer:
                     allreduce_params(runner.model.buffers())
                 self._save_checkpoint(runner)
+                if runner.save_ema_model:
+                    runner.model = backup_model
```
JihwanEom, line 64: Can we move this lines to another hook or stage.py? Only classification task is using mpa checkpoint hook. I think this lines can be an issue if we use official mmclassification.

mpa/modules/hooks/eval_hook.py
```diff
@@ -46,7 +47,11 @@ def after_train_epoch(self, runner):
         if not self.by_epoch or not self.every_n_epochs(runner, self.interval):
             return
         results = single_gpu_test(runner.model, self.dataloader)
-        self.evaluate(runner, results)
+        if hasattr(runner, "ema_model") and (runner.epoch > self.ema_start_eval):
+            results_ema = single_gpu_test(runner.ema_model.module, self.dataloader)
```
JihwanEom, line 51: I'm not sure that `runner.ema_model` always has `module` attribute. Is it also checked for non-distributed environment?

recipes/stages/_base_/data/pipelines/semisl_pipeline.py
```diff
@@ -31,3 +33,14 @@
     dict(type="ToTensor", keys=["gt_label"]),
     dict(type="Collect", keys=["img", "gt_label"]),
 ]
+
+train_rand_pipeline = [
```
JihwanEom: How do you think about separating pipeline file? This file should have only `semisl_pipeline` according to it's filename.

recipes/stages/classification/class_incr.yaml
JihwanEom: But also we can experience -10% performance degradation for edge case in our product. I think seed freezing should be used only for experiment purpose, not for product. It will be better relying the expected value for random seed instead of fixed performance gain or degradation.

recipes/stages/classification/class_incr_multilabel.yaml
```diff
@@ -0,0 +1,37 @@
+_base_: [
```
JihwanEom, line 1: I couldn't find where this file is used. Is this file used only for experiment purpose?

JihwanEom (changes requested):

recipes/stages/classification/class_incr_multilabel.yaml
kprokofi: please, refer to https://github.com/openvinotoolkit/training_extensions/pull/1288
these PRs are connected

recipes/stages/_base_/data/pipelines/semisl_pipeline.py
kprokofi: We can delete it
Last experiments showed that it is not so useful to use RandAugment

mpa/modules/hooks/eval_hook.py
kprokofi: When we run training through OTE CLI, it is always MMDataParallel even though we have training on 1 GPU. For current OTE/SC we don't have distributed environment support through DistributedDataParallel as we don't have multiple GPU training support.

Anyway module attribute it is the EMA attribute itself, not the data parallel attribute.
image

mpa/modules/hooks/eval_hook.py
kprokofi: I don't understand what you mean. Could you please, clarify about parsing the performance? What do you mean by evaluate 'twice'?

mpa/cls/trainer.py
kprokofi: Unfortunately, there is no such attribute in the "before run" stage. So, there are two options: 1) add some not pretty lines to the trainer for counting the max number of images and do not create an EMA model if the specified criteria are not valid 2) Create an EMA model always and check the criteria in the after_train_iter stage and also we need to modify somehow after_train_epoch in eval_hook.

I prefer 1st option. But you can propose another solution

mpa/modules/hooks/checkpoint_hook.py
kprokofi: Is EMA used in other task? If not - it is not the issue. 
If we want to enable EMA in other tasks - we should add this functionality in other way. We need to modify saving checkpoints function for each task. 
For now it works only with classification and doesn't affect other task since it won't have "ema_model" attribute

I propose to leave it as is for now. Because I can't see where we can simply put model substitution for every task.

recipes/stages/_base_/data/pipelines/semisl_pipeline.py
kprokofi: But we use other attributes from these file in templates. But templates are not semi supervised :)
image

recipes/stages/classification/class_incr.yaml
kprokofi: with random seed we also will experience +- 10% in some cases. Moreover, from customer point of view it is not good when you rerun training and obtain different metric. When we have small data it could be a lot different.
@goodsong81 , what do you think about seeds freezing?

mpa/cls/trainer.py
JihwanEom: Hmm, here is some proposal and modification for yours
1) If another hook or algorithm should be aware about the number of images, current code can be nice for us. But as far as I know, there is no hook or algorithm that should know number of images before training so current code is being only for EMA hook. The fact "only for EMA hook" is not problem for me but we can choose another method like 2), without some modification in trainer.py

2) It's more reasonable for me, if we can update EMA hook instead of EvalHook. My imagine is here :
In EMA hook,
` before_train_epoch(self, runner):`
` if not hasattr(runner.number_of_images):`
` runner.number_of_images = len(runner.data_loader.dataset)`
I believe we can be able to avoid checking the criteria in every epoch or iteration.

mpa/modules/hooks/checkpoint_hook.py
JihwanEom: We are not using EMA hook for another task but I agree with you, let's leave it as is.

mpa/modules/hooks/eval_hook.py
JihwanEom: Sorry for my bad explanation, I just asked how about parsing both of them in runner.buffer. Then we will be able to monitor the performance of each model like below example.
`airplane_acc: 70.20 # original model performance`
`EMA_airplane_acc: 72.44 # EMA model performance`
Current implementation is putting only the best result.

kprokofi: done

mpa/cls/trainer.py
kprokofi: I got rid of any changes in trainer.py and transfer it to EMA hook `before_train_epoch` method. Also, I got rid of creating EMA model in case of criteria of the dataset length are not fulfilled

recipes/stages/classification/class_incr_multilabel.yaml
```diff
@@ -0,0 +1,37 @@
...
+evaluation:
+    metric: ['accuracy', 'class_accuracy']
+
+task_adapt:
+    type: 'mpa'
+    op: 'REPLACE'
+
+custom_hooks:
+- type: ModelEmaV2Hook
+- type: LazyEarlyStoppingHook
```
JihwanEom: Please remove early-stop related configs. Early stop is working as learning parameters, so we can handle early stop in template.yaml
More details: https://github.com/openvinotoolkit/training_extensions/blob/develop/external/model-preparation-algorithm/mpa_tasks/apis/task.py#L205

JihwanEom (approved):
LGTM

supersoob (approved):

sovrasov (approved):

PR #69, opened 2022-10-18
Feature/val batch and seed

sungmanc:
* Enable multi-batch validation for detection: reduce the validation time
* Added 'default.py' for classification: apply seed=5, deterministic=False as a default

### DIFF
--- a/mpa/det/trainer.py
+++ b/mpa/det/trainer.py
@@ -69,7 +69,7 @@ class DetectionTrainer(DetectionStage):
 
         # Data
         datasets = [build_dataset(cfg.data.train)]
-        cfg.data.val.samples_per_gpu = cfg.data.samples_per_gpu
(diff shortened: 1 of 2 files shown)


mpa/det/trainer.py
```diff
@@ -69,6 +69,7 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
 
         # Data
         datasets = [build_dataset(cfg.data.train)]
+        cfg.data.val.samples_per_gpu = cfg.data.samples_per_gpu
```
harimkang: I recommend the code below, what do you think?
```
cfg.data.val.samples_per_gpu = cfg.data.get('samples_per_gpu', 1)
```
Also, will this code have any problems with the cpu?

sungmanc: Agree with this change, I changed it as below.
image

In my understanding, changing batch_size can't affect the CPU/GPU training.

harimkang (approved):
LGTM

PR #70, opened 2022-10-19
Add workaround fix when batch size > 1 for MaskRCNN

eugene123tw:
index out of range bug in instance seg after introducing batch size > 1 in this [PR](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/69). For some reason, `scale_factors` in `simple_test_mask` is hardcoded and strictly has only one item in the list  , I'm guessing it's related to OpenVINO export and need some time to investigate

### DIFF
--- a/mpa/det/trainer.py
+++ b/mpa/det/trainer.py
@@ -71,6 +71,12 @@ class DetectionTrainer(DetectionStage):
         datasets = [build_dataset(cfg.data.train)]
         cfg.data.val.samples_per_gpu = cfg.data.get('samples_per_gpu', 1)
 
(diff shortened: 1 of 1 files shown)


sungmanc:
@harimkang @eugene123tw , how about merging this PR as a quick solution? As Eugene said, we need some time to investigate for the instance segmentation.

harimkang:
> @harimkang @eugene123tw , how about merging this PR as a quick solution? As Eugene said, we need some time to investigate for the instance segmentation.

Yes, first of all, I agree.

harimkang (approved):

PR #71, opened 2022-10-19
Kp/devide runners

kprokofi:
The division of the runners proposed to have more flexibility and changeability for different tasks.
The problem was that the 3 tasks used 1 runner configs with 3 different runner classes with the same name. This is very bad. These runner classes substitutes each other in registry. It is now a problem since the runner has been changed for the classification.
I reorganized configs for the runners.

Related OTE PR:
https://github.com/openvinotoolkit/training_extensions/pull/1306


harimkang (commented):
I'm not sure about this PR. If it is to solve the problem of classification, can't we just modify the classification first? (Leave detection and segmentation as it is..?)

In my opinion, this proposal can be a good design or a bad design. (Because there may be duplicate codes) What do you think?

I also think this could be something to consider in our ongoing (or upcoming) refactoring.

supersoob (commented):
I agree with Harim. Could you just seperate clssification runner only?

kprokofi:
> 

Because now we have 3 different runners registered in the registry with the same name and code. In different tasks runners could be used from another task. For example in detection - from mmseg, in classification - from mmdet. Do you think is it a good design? We have a bad dependency and we need to separate runners for the tasks.
I can change it only for the classification - but this problem will remain for other tasks

supersoob (approved):

harimkang:
> > 
> 
> Because now we have 3 different runners registered in the registry with the same name and code. In different tasks runners could be used from another task. For example in detection - from mmseg, in classification - from mmdet. Do you think is it a good design? We have a bad dependency and we need to separate runners for the tasks. I can change it only for the classification - but this problem will remain for other tasks

Sorry for the late reply. I guess I missed it.
Yes this is problematic and I agree.
But if that's the case, `feature/otx` can solve it (not `develop`, I didn't know what effect it would have. Because OTE-CI did not work at that time...). In `feature/otx`, the same code will be organized into a new module called `common`. we can override what we need in each task and use it. So I wanted to solve this in a new refactoring design rather than solve it in develop now.

PR #78, opened 2022-11-10
[RELEASE] Hot-fix drop_last to cls task

JihwanEom:


mpa/cls/trainer.py
```diff
@@ -149,7 +149,12 @@ def train_worker(gpu, dataset, cfg, distributed, validate, timestamp, meta):
         # prepare data loaders
         dataset = dataset if isinstance(dataset, (list, tuple)) else [dataset]
         train_data_cfg = Stage.get_train_data_cfg(cfg)
-        drop_last = train_data_cfg.drop_last if train_data_cfg.get('drop_last', False) else False
+        ote_dataset = train_data_cfg.get('ote_dataset', None)
+        drop_last = False
+        dataset_len = len(ote_dataset) if ote_dataset else 0
+        # if task == h-label & dataset size is bigger than batch size
+        if train_data_cfg.get('hierarchical_info', None) and dataset_len > cfg.data.get('samples_per_gpu', 2):
+            drop_last = True
```
goodsong81, line 157: Is this the right sum of below 2?
image

JihwanEom: Yes, they give same results.

current implementation:
set drop_last=True for h-label, but if sample size is smaller than batch size, set drop_last=False. 
(multi-class/multi-label tasks: set drop_last=False as default, if dataset size is smaller than batch size, maintains False.)

proposed implementation:
set drop_last=False for all task, buf if sample size if bigger than batch size, set drop_last=True only for h-label. 
(multi-class/multi-label tasks: set drop_last=False, we don't check anymore.)

goodsong81 (approved):
LGTM. Please merge this PR after checking test results.

harimkang:
Instance-segmentation:
image

JihwanEom:
Multi-class: 
image

Multi-label:
image

H-label 
image

Detection:
image

harimkang (approved):

PR #83, opened 2022-11-14
Don't apply activations on export in classification

sovrasov:
The motivation is that d-o-r, POT, OpenVINO task and Model API classification components don't expect activated output from IR, so we had wrong confidence results of classification and simply wrong results of multilabel classification in OV tasks / POT. As soon as we remove d-o-r and related code, we can use activated output on the export sage again.


goodsong81 (approved):
LGTM. Let's align outputs after DOR/MPA/OTX consolidation.

JihwanEom (approved):
LGTM

PR #86, opened 2022-11-15
[OTX/Semi-SL] Support unlabeled data via OTX CLI

supersoob:
**This PR includes:**

- Make data_cfg to load and enable semi-sl for classification.
- Minor change to enable semi-sl yaml


mpa/cls/trainer.py
```diff
@@ -65,7 +65,9 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
                     dash_line)
 
         # Data
-        if 'unlabeled' in cfg.data:
+        # FIXME[Soobee] : if train type is not specified, it raises an error in default INCREMENTAL mode.
+        # Should move and refactor in stage.py importing otx.api module
+        if cfg.train_type == 'SEMISUPERVISED':
```
harimkang, line 70: Is this because cfg.data always has unlabeled?

harimkang (commented):
Does this PR only work for classification? Or do all tasks work?

recipes/stages/_base_/data/cls_semi_data.py
```diff
@@ -0,0 +1,30 @@
...
+__test_pipeline = {{_base_.test_pipeline}}
+
+__samples_per_gpu = 32
+
+data = dict(
+    samples_per_gpu=__samples_per_gpu,
+    workers_per_gpu=2,
+    train=dict(
+        type=__dataset_type,
+        pipeline=__train_pipeline),
```
JihwanEom, line 18: Hmm, as far as I know, it should be revised to `train_pipeline_strong`. Am I right @harimkang?

supersoob: I just made a temp data cfg file to enable semi-recipe, because it was not importing(inheriting) any data cfg inside semi recipe file. I expect you to revise it when you start implementation.

mpa/cls/trainer.py
supersoob: No, we don't always have unlabeled. cfg.data cannot have unlabeled when its dataset item entity is empty. but as you said, even user specified 'unlabeled-root-data', we should run INCREMENTAL otherwise train_type is exactly specified as SEMISUPERVISED. In that case, 'unlabeled' would be inside cfg.data even if train_type is INCREMENTAL.

recipes/stages/_base_/data/cls_semi_data.py
JihwanEom: Okay, I'll do that :)

PR #91, opened 2022-11-23
Replace only the first occurrence in the state_dict keys

### LINKED ISSUE #90: `state_dict_hook` might corrupt the model when saving
https://github.com/openvinotoolkit/model_preparation_algorithm/blob/55af1edb6c48ba85b58d01bf1b5ddc9261dbf502/mpa/modules/models/classifiers/sam_classifier.py#L106-L137

Here, we're looking for keys that start with `backbone` or `head`, and then erasing them.
The problem is that we are not only erasing the first one, but all instances.
I had a model with an auxiliary head, named `aux_head` so this happened:

`head.aux_head.weight` → `aux_weight`

We should use `k.replace('backbone.', 1)` and `k.replace('head.', 1)` to make sure we only replace the first appearance.

If at some point we update to Python 3.9 we could even use [`k = k.remove_prefix('backbone.')`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix).

arrufat:
Doing otherwise might corrupt the model.

Fixes #90.

### DIFF
--- a/mpa/modules/models/classifiers/sam_classifier.py
+++ b/mpa/modules/models/classifiers/sam_classifier.py
@@ -106,9 +106,9 @@ class SAMImageClassifier(ImageClassifier):
         if backbone_type == 'OTEMobileNetV3':
             for k, v in state_dict.items():
                 if k.startswith('backbone'):
-                    k = k.replace('backbone.', '')
+                    k = k.replace('backbone.', '', 1)
                 elif k.startswith('head'):
-                    k = k.replace('head.', '')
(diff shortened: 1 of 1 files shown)


harimkang (approved):

arrufat:
Also, I would like this change in the [otx](https://github.com/openvinotoolkit/model_preparation_algorithm/tree/otx) branch, which is the one I am using, should we cherry-pick it, or @supersoob will rebase?

JihwanEom (approved):
LGTM

PR #93, opened 2022-11-25
[OTX] Enable multi-GPU training

eunwoosh:
### Summary
Enable multi-GPU training for classification, detection, segmentation tasks.

### This PR includes

- enable multi GPU training
- combine `train_worker()` into `run()` in each task
- modify some samplers to run on multi GPU training

### DIFF
--- a/mpa/cls/stage.py
+++ b/mpa/cls/stage.py
@@ -8,10 +8,12 @@ import numpy as np
 
 from mmcv import ConfigDict
 from mmcv import build_from_cfg
+from mmcv.parallel import MMDataParallel, MMDistributedDataParallel
 
 from mpa.stage import Stage
 from mpa.utils.config_utils import update_or_add_custom_hook, recursively_update_cfg
 from mpa.utils.logger import get_logger
+from mpa.utils.data_cpu import MMDataCPU
 
 logger = get_logger()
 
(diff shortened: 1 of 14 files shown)


mpa/cls/trainer.py
```diff
@@ -256,37 +192,52 @@ def train_worker(gpu, dataset, cfg, distributed, validate, timestamp, meta):
...
             runner.register_hook(eval_hook(val_dataloader, **eval_cfg), priority='ABOVE_NORMAL')
 
         if cfg.get('resume_from', False):
             runner.resume(cfg.resume_from)
         elif cfg.get('load_from', False):
-            if gpu is None:
-                runner.load_checkpoint(cfg.load_from)
-            else:
-                runner.load_checkpoint(cfg.load_from, map_location=f'cuda:{gpu}')
+            runner.load_checkpoint(cfg.load_from)
```
supersoob: Does it find map_location automatically even when we don't specify the gpu or cpu?

recipes/stages/classification/multilabel/train.yaml
```diff
@@ -1,4 +1,5 @@
 _base_: [
+    '../../_base_/default.py',
```
harimkang, line 2: What is the role of default.py?

mpa/cls/stage.py
```diff
@@ -299,6 +301,26 @@ def configure_hyperparams(cfg, training, **kwargs):
             if lr is not None:
                 cfg.optimizer.lr = lr
 
+    def _put_model_on_gpu(self, model, cfg):
```
harimkang, line 305: Is this function only used for classification? I wonder if it can be used for other tasks as well.

harimkang (commented):
add minor comments

recipes/stages/classification/multilabel/train.yaml
eunwoosh: @supersoob  Could you reply this?

mpa/cls/stage.py
eunwoosh: No, it can't. In other tasks, similar things are done in each framework (mmdet, mmseg).

recipes/stages/classification/multilabel/train.yaml
supersoob: default imports dist.py. multigpu requires this file.

supersoob (approved):
LGTM.

harimkang (approved):

PR #94, opened 2022-11-28
[Semi-SL/Detection] Refactoring mpa/det/ to support multiple learning technique

jaegukhyun:
This PR includes refactoring mpa/det codes.

Current mpa/det's codes highly focus on class incremental learning, and this make hard to enable other training method.
To solve this issue, I split configure function in mpa/det/stage.py into two things. First is configuring configs for general fine tuning. Second is configuring configs for class-incremental learning. 
And I created ClassIncrStage, which inherit base Stage, to support class incremental learning. Now trainer at mpa/det inherits from ClassIncrStage, because class incremental learning is our default recipe.  
To enable semi-sl for detection, SemiStage has to be different with ClassIncrStage. Therefore I created SemiStage. SemiStage inherits from ClassIncrStage, and it modifies some configuration function for supporting semi-sl. Since it inherits ClassIncrStage, it can support class incremental scenarios.
As SemiStage is newly generated, Trainer and Inferencer for semi-sl is also generated, and they inherits SemiStage.

You can test this with otx branch for semi-sl(https://github.com/openvinotoolkit/training_extensions/tree/feature/otx-semisl-det). From this PR link(https://github.com/openvinotoolkit/training_extensions/pull/1398), you can find a description for this otx branch


mpa/det/incr_stage.py
```diff
@@ -0,0 +1,282 @@
+# Copyright (C) 2022 Intel Corporation
```
supersoob: It would be okay to rename this to stage.py because we have seperated folder for semi.

mpa/det/semisl/semisl_stage.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (C) 2022 Intel Corporation
```
supersoob: Same here.

recipes/stages/detection/finetune.py
```diff
@@ -4,7 +4,7 @@
     '../_base_/models/detectors/detector.py'
 ]
 
-model = dict(super_type='UnbiasedTeacher')  # Used as general framework
+model = dict(super_type='UnbiasedTeacher', pseudo_conf_thresh=0.25)  # Used as general framework
```
supersoob: I'm a little worried about if this model config is proper to be inside finetune.py. If it's just overridden in semisl.py, I think it's proper to move there, but I would still wonder if it's right place to be, because recipe file is not for model cfg file. I wanna suggest to move this config to model.py inside semi/model.py in otx. If looks hard, I can agree to leave it in recipe file.

mpa/det/incr_stage.py
jaegukhyun: Incr_stage inherits base stage which is defined in stage.py Therefore it should be named different with stage.py

recipes/stages/detection/finetune.py
jaegukhyun: I understand what you're worrying about. Let me think about it

jaegukhyun: I moved model config to otx side. Could you check it?

jaegukhyun:
@supersoob @sovrasov @kprokofi @JihwanEom It's time to merge this PR, please review this PR. I want you to add your implementation for semi-sl from this branch. After merging this PR, I will open OTX PR, because semi-sl cli test requires this mpa branch

mpa/modules/models/detectors/unbiased_teacher.py
```diff
@@ -128,9 +128,45 @@ def forward_train(self,
...
+        # self.save_pseudo_labels(ul_img0, pseudo_bboxes, pseudo_labels)
 
         return losses
 
+    def save_pseudo_labels(self, ul_imgs, pseudo_bboxes, pseudo_labels):
+        """This saves unlabeled image with pseudo bboxes, this is for debug."""
+        import os
+        import cv2 as cv
+
+        self.save_dir = "/home/jaeguk/workspace/tmp/semisl-det/"
```
supersoob: I want you to revise this save_dir path if you wanna remain this save_pseudo_labels methods.

jaegukhyun: Change the path to /tmp/ Thanks for review!

supersoob (approved):
Overall, it looks good to me. Let's wait for other's reviews. After merging this PR, I think it's gonna be good to make each PR for each task, referring this PR's code change. Thanks for your hard work, Jaeguk:)

goodsong81 (commented):
I support @supersoob's point.
Why don't you create classincr directory?
Being a default option has nothing to do with class/directory structure.
Anyway, it's just a suggestion.

mpa/det/inferrer.py
```diff
@@ -175,9 +175,6 @@ def dummy_dump_saliency_hook(model, input, out):
         feature_vector_hook = dump_features_hook if dump_features else dummy_dump_features_hook
         saliency_map_hook = dump_saliency_hook if dump_saliency_map else dummy_dump_saliency_hook
 
-        # Use a single gpu for testing. Set in both mm_val_dataloader and eval_model
```
JihwanEom, line 178: Could you tell me why you removed these lines?

mpa/det/semisl/trainer.py
```diff
@@ -0,0 +1,169 @@
...
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
+from mmcv.utils import get_git_hash
+from mmdet import __version__
+from mmdet.apis import train_detector
+from mmdet.datasets import build_dataset
+from mmdet.models import build_detector
+from mmdet.utils import collect_env
+# TODO[JAEGUK]: Remove import detection_tasks
```
JihwanEom, line 20: Can we remove now?

mpa/modules/hooks/model_ema_hook.py
```diff
@@ -125,6 +125,15 @@ def _copy_model(self):
 
     def _ema_model(self):
         momentum = min(self.momentum, 1.0)
+
+        # For debug
```
JihwanEom: Do we need debug code?

mpa/modules/models/detectors/unbiased_teacher.py
```diff
@@ -128,9 +128,45 @@ def forward_train(self,
                 loss*self.unlabeled_loss_weight for loss in ul_loss
             ]
         # TODO: apply loss_bbox when adopting QFL;
+        # For debug
```
JihwanEom: Same with here

mpa/modules/models/detectors/unbiased_teacher.py
```diff
@@ -128,9 +128,45 @@ def forward_train(self,
...
+        # self.save_pseudo_labels(ul_img0, pseudo_bboxes, pseudo_labels)
 
         return losses
 
+    def save_pseudo_labels(self, ul_imgs, pseudo_bboxes, pseudo_labels):
```
JihwanEom: Also, looks like only for debug purpose

recipes/stages/detection/semisl.py
```diff
@@ -1,9 +1,13 @@
...
+    '../_base_/data/coco_ubt.py',
+    '../_base_/models/detectors/detector.py'
 ]
 
-model = dict(
-    unlabeled_loss_weight=1.0,
+task_adapt = dict(
+    type='mpa',
+    op='REPLACE',
+    efficient_mode=False,
```
JihwanEom, line 10: Is False default option for detection?

jaegukhyun: I guess so, I copied from class-incr recipe

mpa/modules/hooks/model_ema_hook.py
jaegukhyun: I'll remove this

mpa/det/semisl/trainer.py
jaegukhyun: No, It will be removed when mpa is merged with otx or Inhyuk's PR

mpa/det/inferrer.py
jaegukhyun: I think it is unnecessary code, since model is not used or called after this code

mpa/modules/models/detectors/unbiased_teacher.py
jaegukhyun: I removed all debugging code.

JihwanEom (approved):
LGTM

supersoob (approved):
LGTM

PR #95, opened 2022-11-29
OpenVINO IR module

cih9088:
### Related OTX [PR](https://github.com/openvinotoolkit/training_extensions/pull/1399)
- OpenVINO IR to pytorch graph conversion
- OpenModelZoo downloader

**Need to add test code**

### DIFF
--- a/mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
+++ b/mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
@@ -13,13 +13,24 @@ from ..parser import parameter_parser
 logger = get_logger()
 
 
+NECK_INPUT_TYPES = ["ReduceMean", "MaxPool", "AvgPool"]
+NECK_TYPES = [
+    "Reshape",
+    "Squeeze",
+    "Unsqueeze",
+    "Concat",
+    "Convert",
+    "ShapeOf",
+    "StridedSlice",
+    "Transpose",
+]
+
+
 @PARSERS.register()
 def cls_base_parser(
     graph, component: str = "backbone"
 ) -> Optional[Dict[str, List[str]]]:
     assert component in ["backbone", "neck", "head"]
-
-    possible_neck_input_types = ["ReduceMean", "MaxPool", "AvgPool"]
 
     result_nodes = graph.get_nodes_by_types(["Result"])
     if len(result_nodes) != 1:
@@ -29,7 +40,7 @@ def cls_base_parser(
 
     neck_input = None
     for _, node_to in graph.bfs(result_node, True, 20):
-        if node_to.type in possible_neck_input_types:
+        if node_to.type in NECK_INPUT_TYPES:
             logger.debug(f"Found neck_input: {node_to.name}")
             neck_input = node_to
             break
@@ -38,21 +49,11 @@ def cls_base_parser(
         logger.debug("Can not determine the output of backbone.")
         return None
 
-    part_of_neck_types = [
-        "Reshape",
-        "Squeeze",
-        "Unsqueeze",
-        "Concat",
-        "Convert",
-        "ShapeOf",
-        "StridedSlice",
-        "Transpose",
-    ]
     neck_output = neck_input
     for node_from, node_to in graph.bfs(neck_input, False, 10):
         done = False
         for node_to_ in node_to:
-            if node_to_.type not in part_of_neck_types:
+            if node_to_.type not in NECK_TYPES:
                 done = True
                 break
         neck_output = node_from
--- a/mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
+++ b/mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
@@ -1,0 +1,69 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+from typing import Dict, List, Optional
+
+from mpa.utils.logger import get_logger
+
+from ..parser import parameter_parser
+from ..builder import PARSERS
+
+logger = get_logger()
+
+
+@PARSERS.register()
+def cls_base_parser(
+    graph, component: str = "backbone"
+) -> Optional[Dict[str, List[str]]]:
+    assert component in ["backbone", "neck"]
+
+    possible_neck_input_types = ["ReduceMean", "MaxPool", "AvgPool"]
+
+    result_nodes = graph.get_nodes_by_types(["Result"])
+    if len(result_nodes) != 1:
+        logger.debug("More than one reulst nodes are found.")
+        return None
+    result_node = result_nodes[0]
+
+    neck_input = None
+    for _, node_to in graph.bfs(result_node, True, 20):
+        if node_to.type in possible_neck_input_types:
+            logger.debug(f"Found neck_input: {node_to.name}")
+            neck_input = node_to
+            break
+
+    if neck_input is None:
+        logger.debug("Can not determine the output of backbone.")
+        return None
+
+    if component == "backbone":
+        outputs = [
+            node.name for node in graph.predecessors(neck_input) if node.type != "Constant"
+        ]
+        if len(outputs) != 1:
+            logger.debug(f"neck_input {neck_input.name} has more than one predecessors.")
+            return None
+
+        inputs = parameter_parser(graph)
+        if len(inputs) != 1:
+            logger.debug("More than on parameter nodes are found.")
+            return None
+
+        return dict(
+            inputs=inputs,
+            outputs=outputs,
+        )
+
+    elif component == "neck":
+        part_of_neck_types = ["Reshape"]
+        neck_output = neck_input
+        for node in graph.successors(neck_input):
+            if node.type not in part_of_neck_types:
+                break
+            else:
+                neck_output = node
+        return dict(
+            inputs=[neck_input.name],
+            outputs=[neck_output.name],
+        )
--- a/mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
+++ b/mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
@@ -16,7 +16,7 @@ logger = get_logger()
 def cls_base_parser(
     graph, component: str = "backbone"
 ) -> Optional[Dict[str, List[str]]]:
-    assert component in ["backbone", "neck"]
+    assert component in ["backbone", "neck", "head"]
 
     possible_neck_input_types = ["ReduceMean", "MaxPool", "AvgPool"]
 
@@ -37,6 +37,18 @@ def cls_base_parser(
         logger.debug("Can not determine the output of backbone.")
         return None
 
+    part_of_neck_types = ["Reshape", "Squeeze", "Unsqueeze", "Concat", "Convert", "ShapeOf", "StridedSlice"]
+    neck_output = neck_input
+    for node_from, node_to in graph.bfs(neck_input, False, 10):
+        done = False
+        for node_to_ in node_to:
+            if node_to_.type not in part_of_neck_types:
+                done = True
+                break
+        neck_output = node_from
+        if done:
+            break
+
     if component == "backbone":
         outputs = [
             node.name for node in graph.predecessors(neck_input) if node.type != "Constant"
@@ -56,14 +68,33 @@ def cls_base_parser(
         )
 
     elif component == "neck":
-        part_of_neck_types = ["Reshape"]
-        neck_output = neck_input
-        for node in graph.successors(neck_input):
-            if node.type not in part_of_neck_types:
-                break
-            else:
-                neck_output = node
         return dict(
             inputs=[neck_input.name],
             outputs=[neck_output.name],
+        )
+
+    elif component == "head":
+        inputs = list(graph.successors(neck_output))
+        if len(inputs) != 1:
+            logger.debug(f"neck_output {neck_output.name} has more than one successors.")
+            return None
+
+        outputs = graph.get_nodes_by_types(["Result"])
+        if len(outputs) != 1:
+            logger.debug("more than one network output are found.")
+            return None
+        for node_from, node_to in graph.bfs(outputs[0], True, 5):
+            if node_to.type == "Softmax":
+                outputs = [node_from]
+                break
+
+        if not graph.has_path(inputs[0], outputs[0]):
+            logger.debug(
+                f"input({inputs[0].name}) and output({outputs[0].name}) are reversed"
+            )
+            return None
+
+        return dict(
+            inputs=[input.name for input in inputs],
+            outputs=[output.name for output in outputs],
         )
--- a/mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
+++ b/mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
@@ -6,8 +6,9 @@ from typing import Dict, List, Optional
 
 from mpa.utils.logger import get_logger
 
-from ..parser import parameter_parser
 from ..builder import PARSERS
+from ..parser import parameter_parser
+
 
 logger = get_logger()
 
@@ -37,7 +38,16 @@ def cls_base_parser(
         logger.debug("Can not determine the output of backbone.")
         return None
 
-    part_of_neck_types = ["Reshape", "Squeeze", "Unsqueeze", "Concat", "Convert", "ShapeOf", "StridedSlice"]
+    part_of_neck_types = [
+        "Reshape",
+        "Squeeze",
+        "Unsqueeze",
+        "Concat",
+        "Convert",
+        "ShapeOf",
+        "StridedSlice",
+        "Transpose",
+    ]
     neck_output = neck_input
     for node_from, node_to in graph.bfs(neck_input, False, 10):
         done = False
@@ -51,10 +61,14 @@ def cls_base_parser(
 
     if component == "backbone":
         outputs = [
-            node.name for node in graph.predecessors(neck_input) if node.type != "Constant"
+            node.name
+            for node in graph.predecessors(neck_input)
+            if node.type != "Constant"
         ]
         if len(outputs) != 1:
-            logger.debug(f"neck_input {neck_input.name} has more than one predecessors.")
+            logger.debug(
+                f"neck_input {neck_input.name} has more than one predecessors."
+            )
             return None
 
         inputs = parameter_parser(graph)
@@ -75,9 +89,9 @@ def cls_base_parser(
 
     elif component == "head":
         inputs = list(graph.successors(neck_output))
-        if len(inputs) != 1:
-            logger.debug(f"neck_output {neck_output.name} has more than one successors.")
-            return None
+        #  if len(inputs) != 1:
+        #      logger.debug(f"neck_output {neck_output.name} has more than one successors.")
+        #      return None
 
         outputs = graph.get_nodes_by_types(["Result"])
         if len(outputs) != 1:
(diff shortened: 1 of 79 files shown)


mpa/modules/ov/models/mmcls/heads/cls_head.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+from mmcls.models.builder import HEADS
+from mmcls.models.heads import ClsHead as OriginClsHead
+
+
+@HEADS.register_module()
+class ClsHead(OriginClsHead):
```
harimkang, line 10: This conflicts with the Head of the existing mmcls. we need to change the name
Or if there is no problem, we should `force=True`.

mpa/modules/ov/models/mmdet/necks/mmov_ssd_neck.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+from typing import Dict, List, Optional, Union
+
+import torch
+from mmdet.models.builder import NECKS
+from mmdet.models.necks.ssd_neck import SSDNeck
```
harimkang: The `mmdet` currently used in otx cannot be imported. If this is supported before mmpublic-det integration, it seems that this should be corrected.

mpa/modules/ov/models/mmdet/roi_heads/roi_extractors/single_level_roi_extractor.py
```diff
@@ -0,0 +1,64 @@
...
+from mmdet.models.roi_heads.roi_extractors.single_level_roi_extractor import (
+    SingleRoIExtractor as OriginSingleRoIExtractor,
+)
+from torch import nn
+from torch.nn import functional as F
+from torch.nn.modules.utils import _pair
+
+
+@ROI_EXTRACTORS.register_module()
+class SingleRoIExtractor(OriginSingleRoIExtractor):
```
harimkang, line 16: This class also conflicts with Extractor registered in the existing registry.

harimkang (commented):
Based on [this PR](https://github.com/openvinotoolkit/training_extensions/pull/1399), I tried running sample with classification after reinstalling mpa as a branch of the current PR.
(otx/algorithms/init_venv.sh + mpa reinstalling as this PR)

I left a comment first on only the things needed to run the sample from current feature/otx

mpa/modules/ov/models/mmcls/heads/cls_head.py
cih9088: Since this class merely adds negligible features, I believe `force=True` would be enough.
If you insisted to rename it, I would. Let me know.

mpa/modules/ov/models/mmdet/roi_heads/roi_extractors/single_level_roi_extractor.py
cih9088: Since this class merely adds negligible features, I believe `force=True` would be enough.
If you insisted to rename it, I would. Let me know.

mpa/modules/ov/models/mmcls/heads/cls_head.py
harimkang: I also think it's best to do `force=True` if this class doesn't have a big impact on the existing ones.

mpa/modules/ov/models/mmdet/necks/mmov_ssd_neck.py
cih9088: Indeed, I assumed public `mmdet`. I would copy `SSDNeck` from public `mmdet` for the time being.

mpa/modules/ov/models/mmcls/backbones/mmov_backbone.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import List, Dict
+
+from mmcls.models.builder import BACKBONES
+
+from ...mmov_model import MMOVModel
+from ....graph.parsers.cls.cls_base_parser import cls_base_parser
+
+
+@BACKBONES.register_module()
+class MMOVBackbone(MMOVModel):
```
harimkang, line 14: The `MMOVBackbone` class of the same name is implemented for each task. Could it be possible to unify this? 
If merging is not possible, consider renaming the individual classes..
Also, how about registering this as mmcv's backbone?
you can refer this [code](https://github.com/openvinotoolkit/training_extensions/blob/b19ab2f359ce10885d0f602f37890a70e321945e/otx/algorithms/common/adapters/mmcv/models/builder.py#L20),
maybe.. we can use `BACKBONES = Registry("models", parent=MMCV_MODELS, scope="omz")` and then we can use `omz.MMOVBackbone` in model configuration.

cih9088: First of all, we can't unify backbones since each library requires different outputs of backbone.
Could you explain why we need to name them differently?

I haven't seen, `omz.MMOVBackbone`, this kind of use case in OTX. Could you point it out?
What do you mean by **use omz.MMOVBackbone in model configuration.**?

harimkang: Renaming was for my suggestion about registering in one registry of mmcv.

Recently `feature/otx`, It has been modified to register and find backbones in the registry in the form of `otx.OTXMobileNetV3`, `mmdet.ResNet` or `torchvision.resnet`. It was just my suggestion for public backbone support.

However, leave this as it is in this PR for now, and see if I can unify the format. (according to the style of otx)

goodsong81 (commented):
@harimkang's comments are all for ease of otx find & build.
If it does not harm the functionality, let's align to some extent.
Of course, some part could be done in caller part. :)
In addition, @cih9088, we should not assume that "force=True" is forced in mmcv installation as we will remove init_venv.sh eventually.

cih9088:
@harimkang 
After looking into [your PR](https://github.com/openvinotoolkit/training_extensions/pull/1391), I get the idea.
I have no objection to it but I believe registering `MMOVBackbone` to `mmcv`'s backbone should be implemented in OTX not MPA even though renaming should be done in MPA.

I have rebased onto current branch and fixed  bugs in https://github.com/openvinotoolkit/model_preparation_algorithm/pull/95/commits/249baf0852c7c94e9f93632528f44d779c08b792

mpa/modules/ov/models/__init__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (C) 2022 Intel Corporation
```
harimkang, line 1: Why are we trying to import (mmcls/mmdet/mmseg) here?
(Currently, many warning logs appear when importing non-public (otx side)`mmseg`. I don't know exactly why... However, I mentioned it because this part seems to cause more logs.)

cih9088: It was to register each `MMOV` model to a library specific registry.
Could you share warning logs?

harimkang: Sorry for the confusion. I'm not sure if here is wrong. It seems to be importing `nncf.torch` from somewhere.. (Even though nncf isn't working, it looks like it's importing somewhere.)

and, Is this import work to check if import is possible in advance?
Shouldn't this be inside the folders in `mmcls`, `mmdet`, `mmseg` respectively?
For example.. I don't know if it's appropriate to import `mmdet` and `mmseg` once to use classification (This is probably enough for `mmcls`.).

harimkang (approved):
For me, Sample with this PR is working with OMZ model configuration for classification.
And when I checked the otx cli test locally using the corresponding MPA branch, there seemed to be no problem.

After the PR is merged, I plan to create a PR that allows me to configure and run the OMZ model through `otx find & build`, and I plan to add a test to check if it works by OMZ backbones.

@cih9088 This PR only supports classification first, right?

mpa/modules/ov/models/__init__.py
cih9088: I don't mind moving this test import to anywhere else. Your call.

cih9088:
> For me, Sample with this PR is working with OMZ model configuration for classification. And when I checked the otx cli test locally using the corresponding MPA branch, there seemed to be no problem.
> 
> After the PR is merged, I plan to create a PR that allows me to configure and run the OMZ model through `otx find & build`, and I plan to add a test to check if it works by OMZ backbones.
> 
> @cih9088 This PR only supports classification first, right?

Even though some detection models and segmentation models are supported as well but, yes, OMZ download and convert features are only available for classification models for the time being.

harimkang:
@cih9088 Hi. I have one question. Is there a need for different data pipelines for each backbone?

cih9088:
> @cih9088 Hi. I have one question. Is there a need for different data pipelines for each backbone?

Which pipeline? I don't get it. Could you point it out in code and explain it a bit more?

harimkang:
> Which pipeline? I don't get it. Could you point it out in code and explain it a bit more?

For example, when trying to use alexnet, I met the error below
image
and it works only when I change the value for resize to 227.
This size must be matched for each backbone to operate..? (It is correct that it is fixed, and this should be modified by the user now, right?)

cih9088:
I see your problem. Here is a thing.
Some models have hard-coded values related to their input shape so that arbitrary input shape leads a wrong behaviour.
One could change model input to dynamic shapes with provided [OpenVINO API](https://docs.openvino.ai/latest/openvino_docs_OV_UG_DynamicShapes.html), but it has its own drawback.
`dynamic shape` API generates lots of new nodes assigned random name on them.
The number of nodes generated is not a problem. What matters is a random name of each generated nodes.
Because of the random names, a graph constructed each time one initialized `MMOV` series model is not identical so that we can not load `state_dict`, which means, after training, we can not load the model.

goodsong81:
In my opinion, the input resolution should be model or use-case specific so that backbones should work in any resolution. I suppose backbone weights in *.bin files should be free from input resolution logically. On the other hand, value normalization is specific to the backbone, and IRs have their own input normalization layers in general. So we can omit input norm in case of OV backbone.

goodsong81:
> I see your problem. Here is a thing. Some models have hard-coded values related to their input shape so that changing input shape resulting a wrong behaviour. One could change model input to dynamic shapes with provided [OpenVINO API](https://docs.openvino.ai/latest/openvino_docs_OV_UG_DynamicShapes.html), but it has its own drawback. `dynamic shape` API generates lots of new nodes assigned random name on them. The number of nodes generated is not a problem. What matters is a random name of each generated nodes. Because of the random names, a graph constructed each time one initialized `MMOV` series model is not identical so that we can not load `state_dict`, which means, after training, we can not load the model.

IR model needs reshape()-like API support but torch model does not.
Can we create torch model from IR in such way where resolution is adaptive to input?

cih9088:
> > I see your problem. Here is a thing. Some models have hard-coded values related to their input shape so that changing input shape resulting a wrong behaviour. One could change model input to dynamic shapes with provided [OpenVINO API](https://docs.openvino.ai/latest/openvino_docs_OV_UG_DynamicShapes.html), but it has its own drawback. `dynamic shape` API generates lots of new nodes assigned random name on them. The number of nodes generated is not a problem. What matters is a random name of each generated nodes. Because of the random names, a graph constructed each time one initialized `MMOV` series model is not identical so that we can not load `state_dict`, which means, after training, we can not load the model.
> 
> IR model needs reshape()-like API support but torch model does not. Can we create torch model from IR in such way where resolution is adaptive to input?

Let me clarify it a bit further.
IR models from OMZ have a static input shape.
In some of models, the input shape is hard-coded in IR, not depends on input shape so that arbitrary input shape to model would cause a hard to debug error.

One can change this static shape to dynamic shape using [OpenVINO API](https://docs.openvino.ai/latest/openvino_docs_OV_UG_DynamicShapes.html) before converting to pytorch model. I believe the API would change all the hard-coded input shape thing.
However, the problem is that the API generates random named nodes.
This causes a graph mismatch every time you initialize the model, which means we can not load a trained model because a graph of trained model and a graph of newly initialize model are not identical.

Actually, the randomness is not really random. I think there is some sort of session thing in ngraph because each time when one applies `dynamic shape` API after completely quitting the process, generated random names are the same.
The problem arises when one uses `dynamic shape` API multiple times in one process.

goodsong81:
I'm not talking about IR model reshape, but torch model.
(x) reshape IR to (w,h) -> convert to torch model
(o) convert to torch model -> infer on (w,h) input

I understand that IR models have static shape. But torch CNNs are dynamic in spatial resolution.
The question is why dynamic torch model is subject to original IR's static shape.

cih9088:
> I'm not talking about IR model reshape, but torch model. (x) reshape IR to (w,h) -> convert to torch model (o) convert to torch model -> infer on (w,h) input
> 
> I understand that IR models have static shape. But torch CNNs are dynamic in spatial resolution. The question is why dynamic torch model is subject to original IR's static shape.

Converted torch models depend on IR graph. If IR graph has a hard-coded value related to input shape, that would apply to the torch model as well. 
To loosen shape check, we have to verify each IR has no hard-coded values.

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
```diff
@@ -0,0 +1,114 @@
...
+        "Squeeze",
+        "Unsqueeze",
+        "Concat",
+        "Convert",
+        "ShapeOf",
+        "StridedSlice",
+        "Transpose",
+    ]
+    neck_output = neck_input
+    for node_from, node_to in graph.bfs(neck_input, False, 10):
```
wonjuleee, line 53: could you describe what False and 10 means?

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
```diff
@@ -0,0 +1,114 @@
...
+    part_of_neck_types = [
+        "Reshape",
+        "Squeeze",
+        "Unsqueeze",
+        "Concat",
+        "Convert",
+        "ShapeOf",
+        "StridedSlice",
+        "Transpose",
+    ]
```
wonjuleee: Isn't it good to move this to global variable at the top of the code?

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
```diff
@@ -0,0 +1,114 @@
...
+logger = get_logger()
+
+
+@PARSERS.register()
+def cls_base_parser(
+    graph, component: str = "backbone"
+) -> Optional[Dict[str, List[str]]]:
+    assert component in ["backbone", "neck", "head"]
+
+    possible_neck_input_types = ["ReduceMean", "MaxPool", "AvgPool"]
```
wonjuleee: Isn't it good to move this to global variable at the top of the code?

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
wonjuleee: how about renaming this to `possible_neck_types`

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
```diff
@@ -0,0 +1,114 @@
...
+        if not graph.has_path(inputs[0], outputs[0]):
+            logger.debug(
+                f"input({inputs[0].name}) and output({outputs[0].name}) are reversed"
+            )
+            return None
+
+        return dict(
+            inputs=[input.name for input in inputs],
+            outputs=[output.name for output in outputs],
+        )
```
wonjuleee, line 115: please add 
else:
    raise error

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
```diff
@@ -0,0 +1,114 @@
...
+            outputs=[neck_output.name],
+        )
+
+    elif component == "head":
+        inputs = list(graph.successors(neck_output))
+        #  if len(inputs) != 1:
+        #      logger.debug(f"neck_output {neck_output.name} has more than one successors.")
+        #      return None
+
+        outputs = graph.get_nodes_by_types(["Result"])
```
wonjuleee, line 97: Should "Result" be unique for the output layer name?
If not, how about set this to `possible_output_types` as a global variable?

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
cih9088: I made this function output `None` on purpose in case it fails to parse automatically because it is a utility function.
Error would be raised in [here](https://github.com/cih9088/model_preparation_algorithm/blob/feature/ov-otx/mpa/modules/ov/models/mmcls/backbones/mmov_backbone.py#L20-L26) where this function is actually used.

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
cih9088: `def bfs(self, node, reverse, depth_limit)`
The depth limit is set based on OMZ IR.

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
cih9088: `Result` is not a name but a node type and it is a part of head section.
This function tries to find neck component and output backbone, neck, head depending on input parameter.
I believe there would be no other types than `Result`.

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
cih9088: Fixed in https://github.com/openvinotoolkit/model_preparation_algorithm/pull/95/commits/e173066e7d7a43670452fb1b3dc9e99690c4503a

mpa/modules/ov/graph/parsers/cls/cls_base_parser.py
cih9088: Fixed in https://github.com/openvinotoolkit/model_preparation_algorithm/pull/95/commits/e173066e7d7a43670452fb1b3dc9e99690c4503a

cih9088:
@harimkang FYI, you could pass `verify_shape=False` when initializing MMOV model.
For example,
```python
model_cfg = dict(
    type="ImageClassifier",
    backbone=dict(
        type="MMOVBackbone",
        model_path="omz://alexnet",
        remove_normalize=False,
        merge_bn=True,
        paired_bn=True,
        verify_shape=False,
        outputs="fc8/flatten_fc_input",
    ),
    neck=None,
    head=dict(
        type="MMOVClsHead",
        model_path="omz://alexnet",
        inputs="fc8/WithoutBiases",
        outputs="fc8",
        verify_shape=False,
        loss=dict(type="CrossEntropyLoss", loss_weight=1.0),
    ),
)
```

This would not verify that pytorch input shape does match to IR input shape

harimkang:
> @harimkang FYI, you could pass `verify_shape=False` when initializing MMOV model. For example,

thank you, I'll check!

mpa/modules/ov/models/__init__.py
harimkang: image
I checked this again. I confirmed that unnecessary importing (& logs) do not come out by deleting the test import.
Could you please remove this? And how about moving this to the init of each folder (mmcls, mmdet, mmseg)?

cih9088: I can confirm that using public `mmseg` library, the warning messages is not appeared.
image
Do you want me to comment them out for the moment?

cih9088: Commented them out in  https://github.com/openvinotoolkit/model_preparation_algorithm/pull/95/commits/9d4dea34e7d1b146f339a5aefe9d46757238a18a

harimkang:
@cih9088  could you fix the conflicting part?
@goodsong81 Do you agree to merge this PR after the conflict is resolved?

cih9088:
@harimkang Rebased onto current `OTX`

goodsong81 (approved):

PR #96, opened 2022-11-29
Train NNCF under MPA

cih9088:


mpa/det/stage.py
```diff
@@ -124,10 +124,14 @@ def configure_data(self, cfg, training, **kwargs):
...
-                train_cfg.data_classes = cfg.data.train.pop('data_classes', None)
-                train_cfg.new_classes = cfg.data.train.pop('new_classes', None)
+                if cfg.data.train.get('otx_dataset', None) is not None:
+                    train_cfg.otx_dataset = cfg.data.train.pop('otx_dataset')
+                if cfg.data.train.get('labels', None) is not None:
+                    train_cfg.labels = cfg.data.train.get('labels')
+                if cfg.data.train.get('data_classes', None) is not None:
+                    train_cfg.data_classes = cfg.data.train.pop('data_classes')
+                if cfg.data.train.get('new_classes', None) is not None:
+                    train_cfg.new_classes = cfg.data.train.pop('new_classes')
```
goodsong81, line 134: Any issues found in the previous code?

mpa/modules/models/losses/am_softmax_loss_with_ignore.py
```diff
@@ -14,19 +14,19 @@
...
-        from .cross_entropy_loss import NormalizedCrossEntropy
+        from .cross_entropy_loss import NormalizedCrossEntropyLoss
         return NormalizedCrossEntropy(**kwargs)
     elif name == 'rce':
-        from .cross_entropy_loss import ReverseCrossEntropy
+        from .cross_entropy_loss import ReverseCrossEntropyLoss
         return ReverseCrossEntropy(**kwargs)
     elif name == 'sl':
-        from .cross_entropy_loss import SymmetricCrossEntropy
+        from .cross_entropy_loss import SymmetricCrossEntropyLoss
```
goodsong81: Was the "Loss" postfix appended?
Should be applied to line#21,24,27, too?

cih9088: Thanks for catching it. Fixed it in https://github.com/openvinotoolkit/model_preparation_algorithm/pull/96/commits/ba1bbad8ff13b73268edb12fafb8816b14297e14

mpa/det/stage.py
cih9088: Yes, it does.
Sometimes, `train_cfg.otx_dataset`, etc. have been set somewhere, and `cfg.data.train.otx_dataset`, etc. are `None` or empty. We should set `train_cfg.otx_dataset` only if `cfg.data.train.otx_dataset` has some value in it.

cih9088: Why do we need this block of code in the first place? Do we have any use case?
In my opinion, `otx_dataset`, etc. in `cfg.data.train`, which is a wrapper dataset, have to be considered a bug, letting users know you are doing wrong.
I would suggest to get rid of it.

goodsong81: Yeah, I agree. `otx_dataset`, which is actually not a config object but a real object, is kind of hack to feed Geti dataset to mmcv frameworks. If-else logics here need to be refined anyway.
Let's do it it separate PRs.

goodsong81 (approved):

PR #97, opened 2022-11-29
Class-wise saliency map generation for the detection task

negvet:
Enable class-wise saliency map generation for the detection case.

Tested with all detection models from `external\model-preparation-algorithm\configs\detection\` on PASCAL VOC dataset

related PR:
https://github.com/openvinotoolkit/training_extensions/pull/1402

Some examples (mobilenetv2_ssd):
image

### DIFF
--- a/mpa/det/inferrer.py
+++ b/mpa/det/inferrer.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2022 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 #
+from contextlib import nullcontext
 from typing import List, Tuple
 
 import torch
@@ -16,6 +17,7 @@ from mmdet.apis import single_gpu_test
 from mpa.registry import STAGES
 from .stage import DetectionStage
 from mpa.utils.logger import get_logger
+from mpa.modules.hooks.auxiliary_hooks import SaliencyMapHookDet
 
 
 logger = get_logger()
@@ -148,7 +150,6 @@ class DetectionInferrer(DetectionStage):
 
         eval_predictions = []
         feature_vectors = []
-        saliency_maps = []
 
         def dump_features_hook(mod, inp, out):
             with torch.no_grad():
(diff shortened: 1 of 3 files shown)


mpa/det/inferrer.py
```diff
@@ -159,30 +159,15 @@ def dump_features_hook(mod, inp, out):
...
-        saliency_map_hook = dump_saliency_hook if dump_saliency_map else dummy_dump_saliency_hook
 
         # Use a single gpu for testing. Set in both mm_val_dataloader and eval_model
         if is_module_wrapper(model):
             model = model.module
         with eval_model.module.backbone.register_forward_hook(feature_vector_hook):
-            with eval_model.module.backbone.register_forward_hook(saliency_map_hook):
+            with SaliencyMapHookDet(eval_model.module) if dump_saliency_map else nullcontext() as shook:
                 eval_predictions = single_gpu_test(eval_model, data_loader)
+                saliency_maps = shook.records if dump_saliency_map else [None] * len(self.dataset)
```
sungmanc: How about using `saliency_hook` instead of `shook`?

mpa/modules/hooks/auxiliary_hooks.py
```diff
@@ -132,6 +134,111 @@ def __exit__(self, exc_type, exc_value, traceback):
         self._handle.remove()
 
 
+class SaliencyMapHookDet(SaliencyMapHook):
```
sungmanc: `DetSaliencyMapHook` is more attractable to me. Because this is also a kind of `Hook`.

goodsong81 (approved):
LGTM. I agree with @sungmanc for the naming, but not mandatory.

mpa/modules/hooks/auxiliary_hooks.py
negvet: I agree. Renamed

mpa/det/inferrer.py
negvet: Updated!

sungmanc (approved):
Thanks for reflecting comments, LGTM

PR #98, opened 2022-12-05
[XAI] hot-fix of error in Detection XAI support

dongkwan-kim01:
This is the fix of the error issued in [issue](https://github.com/openvinotoolkit/training_extensions/pull/1402#issuecomment-1336631895), which fixes the `simple_test` function in the class`SAMDetectorMixin`.

### DIFF
--- a/mpa/modules/models/detectors/sam_detector_mixin.py
+++ b/mpa/modules/models/detectors/sam_detector_mixin.py
@@ -4,7 +4,7 @@
 
 import torch
 
-from mmdet.models.detectors import BaseDetector, SingleStageDetector
+from mmdet.models.detectors import BaseDetector, TwoStageDetector
 from mmdet.utils.deployment.export_helpers import get_feature_vector, get_saliency_map
 from mmdet.integration.nncf.utils import no_nncf_trace
 from mpa.modules.hooks.auxiliary_hooks import DetSaliencyMapHook
@@ -25,12 +25,13 @@ class SAMDetectorMixin(BaseDetector):
                     proposals=None,
                     rescale=False,
                     postprocess=True):
-
         """
         Class-wise Saliency map for Single-Stage Detector, otherwise use class-ignore saliency map.
         """
-        x = self.extract_feat(img)
-        if isinstance(self, SingleStageDetector):
+        if isinstance(self, TwoStageDetector):
+            return super().simple_test(img, img_metas, proposals, rescale, postprocess)
+        else:
+            x = self.extract_feat(img)
             outs = self.bbox_head(x)
             with no_nncf_trace():
                 bbox_results = \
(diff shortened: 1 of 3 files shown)


sungmanc:
Let's wait the result of OTE CI test, first.

mpa/modules/models/detectors/sam_detector_mixin.py
```diff
@@ -19,22 +19,33 @@ def train_step(self, data, optimizer, **kwargs):
...
-                self.postprocess(det_bboxes, det_labels, None, img_metas, rescale=rescale)
-                for det_bboxes, det_labels in bbox_results
-            ]
-        return bbox_results
+            if postprocess:
+                bbox_results = [
+                    self.postprocess(det_bboxes, det_labels, None, img_metas, rescale=rescale)
+                    for det_bboxes, det_labels in bbox_results
+                ]
+            return bbox_results
```
goodsong81, line 51: I might have missed the code review for this part.
Could you explain why XAI feature is implemented in "SAM" mixin rather than implementing "XAIMixin"?
It's a refactoring issue, not mandatory fix in this PR, though :)

goodsong81 (commented):
So, we are not supporting the updated per-class XAI output for any of 2-stage detectors (including all of our instance segmentation models). Is it correct?
@negvet Do you have a plan for it?

mpa/modules/models/detectors/sam_detector_mixin.py
dongkwan-kim01: I think this would be because the saliency map exporting is subordinate to the inference step. How's your thought @negvet ?

dongkwan-kim01:
> So, we are not supporting the updated per-class XAI output for any of 2-stage detectors (including all of our instance segmentation models). Is it correct? @negvet Do you have a plan for it?

Yes, currently only supports class-wise saliency maps for single-stage detectors because of the structural difference(classifier head is in the second stage in two-stage detectors), or our first plan is for ATSS, SSD, YOLOX which all are single-stage.

goodsong81:
@dongkwan-kim01 We've created a release branch (releases/OTE_0.4.0). Could you duplicate this PR to the release branch and also create another PR for OTE release branch (release/v0.4.0-geti1.1.0)?

dongkwan-kim01:
> @dongkwan-kim01 We've created a release branch (releases/OTE_0.4.0). Could you duplicate this PR to the release branch and also create another PR for OTE release branch (release/v0.4.0-geti1.1.0)?

Yes, I will. Thank you!

mpa/modules/models/detectors/sam_detector_mixin.py
negvet: I agree actually that it make sense to implement additional `Mixin` for XAI purposes (if it is ok for us to have one more inheritance, is it?).

But for the detection I followed the classification convention:
https://github.com/openvinotoolkit/model_preparation_algorithm/blob/a99aea0296a6c1987c014d55c26dfb56ca8be611/mpa/modules/models/classifiers/sam_classifier.py#L274
which have been like this even before default XAI algo replacement.

negvet:
> So, we are not supporting the updated per-class XAI output for any of 2-stage detectors (including all of our instance segmentation models). Is it correct? @negvet Do you have a plan for it?

Sure, we plan to look into 2-stage detectors in Q12023

negvet:
> This is the fix of the error issued in [issue](https://github.com/openvinotoolkit/training_extensions/pull/1402#issuecomment-1336631895), which fixes the `simple_test` function in the class`SAMDetectorMixin`.

@dongkwan-kim01 Thank you for the fast fix!

goodsong81 (approved):

PR #100, opened 2022-12-05
[MPA] Recover Semi-SL for multi-class classification in OTX

JihwanEom:
### Summary
- Enable Semi-SL for multi-class classification
- Separate SemiClsHead into SemiLinearClsHead/SemiNonLinearClsHead
- Refactor SemiSLClassifier to minimize overriding, aligned with updated mmcls

Related PR: https://github.com/openvinotoolkit/training_extensions/pull/1415


mpa/cls/trainer.py
```diff
@@ -65,8 +65,12 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
                     dash_line)
 
         # Data
-        if 'unlabeled' in cfg.data:
+        if 'unlabeled' in cfg.data and cfg.train_type == 'SEMISUPERVISED':
             datasets = [[build_dataset(cfg.data.train), build_dataset(cfg.data.unlabeled)]]
+            datasets[0][0].samples_per_gpu = cfg.data.samples_per_gpu
+            datasets[0][0].workers_per_gpu = cfg.data.workers_per_gpu
+            datasets[0][1].samples_per_gpu = cfg.data.unlabeled.get('samples_per_gpu', cfg.data.samples_per_gpu*3)
```
harimkang: This is likely to cause index errors. As far as I know, there may be no unlabeled dataset.

mpa/cls/__init__.py
```diff
@@ -37,7 +37,8 @@
 import mpa.modules.models.heads.custom_hierarchical_linear_cls_head
 import mpa.modules.models.heads.custom_hierarchical_non_linear_cls_head
 
-import mpa.modules.models.heads.semisl_cls_head
+import mpa.modules.models.heads.semi_linear_cls_head
+import mpa.modules.models.heads.semi_non_linear_cls_head
```
harimkang, line 43: I don't know why it has to be separated into two. Are the results different depending on the head? Or do not support it? As far as I know, `semisl_cls_head` can support all models.

mpa/cls/trainer.py
```diff
@@ -161,8 +165,8 @@ def train_worker(gpu, dataset, cfg, distributed, validate, timestamp, meta):
                 sub_loaders = [
                     build_dataloader(
                         sub_ds,
-                        sub_ds.samples_per_gpu if hasattr(sub_ds, 'samples_per_gpu') else cfg.data.samples_per_gpu,
-                        sub_ds.workers_per_gpu if hasattr(sub_ds, 'workers_per_gpu') else cfg.data.workers_per_gpu,
+                        sub_ds.samples_per_gpu,
```
harimkang: Will all tests pass even if the attr check disappears?

recipes/stages/classification/semisl.yaml
```diff
@@ -20,3 +21,6 @@ custom_hooks:
   priority: ABOVE_NORMAL
 - type: SemiSLClsHook
 
+task_adapt:
```
harimkang, line 24: This seems like something worth discussing.

harimkang (changes requested):
I left a comment because I wasn't sure why the head had to be split into two, and other than that, I left a comment on the part that might cause an error.

mpa/cls/trainer.py
harimkang: + It looks good to be able to receive the unlabeled batch size from otx. (not a fixed 3x value)
It seems inconvenient to have to edit this code to fix this (unlabeled batch size).

mpa/cls/__init__.py
JihwanEom: Yes, `semisl_cls_head` can support all models but if then, for mobilenetv3, there is architecture difference (we can't load model weights from supervised/incremental). To prevent that, I've separated into `semisl_linear_cls_head` and `semisl_non_linear_cls_head`.

recipes/stages/classification/semisl.yaml
JihwanEom: I agree, please share your nice thinks @jaegukhyun , @supersoob , @goodsong81

mpa/cls/__init__.py
harimkang: Does the head have a big impact on the load model? Also, will this change the results? For reference, all of the previous Semi-SL experiments were the results of one head.

mpa/cls/trainer.py
JihwanEom: For supervised/incremental learning, there was no problem. I'll try to catch edge cases for these lines.

mpa/cls/__init__.py
JihwanEom: Yes, significantly affects. (92.84 -> 0.04) when inference with supervised template. I think the output of semi-sl should support inference with supervised template.

mpa/cls/trainer.py
JihwanEom: Yes, please share your ideas if you have good suggestions. Same mechanism can be applied on other tasks, I think. @supersoob, @jaegukhyun

mpa/cls/__init__.py
harimkang: Well, that's right. If that's the case, please tie those two heads together into one file. semi_sl_head.py

JihwanEom: Hmm, but other heads are separated, and I followed the convention. Do you think existing files also should be merged in one file?

harimkang: Don't worry about the other files for now.

JihwanEom: Okay, I'll reflect your comments. Thank you.

mpa/cls/trainer.py
jaegukhyun: We should discuss about the way to build unlabeled dataset, in detection case we use unlabeled data hook

mpa/modules/datasets/composed_dataloader.py
```diff
@@ -46,6 +46,7 @@ def set_epoch(self, epoch):
     def __init__(self, loaders=[]):
         self.loaders = loaders
         self.max_iter = len(self.loaders[0])
+        self.dataset = [i.dataset for i in self.loaders]
```
jaegukhyun: Maybe self.datasets is more propriate name?

mpa/cls/trainer.py
harimkang: I think it is necessary to expose the unlabeled batch size in the otx template. Other tasks may be run later with different settings (different unlabeled batch size).

recipes/stages/_base_/data/cls_semi_data.py
```diff
@@ -0,0 +1,34 @@
+_base_ = [
```
jaegukhyun, line 1: Should it be moved to otx/algo/cls/model/data_pipeline.py?

recipes/stages/classification/semisl.yaml
jaegukhyun: I don't get what is the point of discussing. If it is about adding task_adapt flag to semisl recipe, we agreed on semi-sl should support task_adapt and ignore label.

mpa/modules/datasets/composed_dataloader.py
JihwanEom: I agree, but please refer ema hook. If we set the attribute as self.datasets on ComposedDL, that can be occur some.. complex if statement in ema hook.

recipes/stages/_base_/data/cls_semi_data.py
JihwanEom: Supervised/incremental is not separated with data_pipeline.py and model.py. I think we need anothet PR to align implementation method for Semi-SL include unlabeled_hook in classification. (also for segmentation)

recipes/stages/classification/semisl.yaml
harimkang: As far as I know it is most likely not supported in semi right away. In particular, ignore labels are not supported due to differences in loss.

jaegukhyun (changes requested):
Generally it looks good to me, I left some comments. And I have two question
1. In classification task, does configure function in mpa/cls/stage.py works same in semi-sl?
2. What is the difference between in linear head and non-linear head? Does it is number of layers of head(linear: 1, non-linear: multiple)?

recipes/stages/classification/semisl.yaml
JihwanEom: Yes, I think we can resolve this conversation as-is.

recipes/stages/_base_/data/cls_semi_data.py
harimkang: In my opinion, unlabeled_hook is not needed for classification.

recipes/stages/classification/semisl.yaml
jaegukhyun: Okay, then @JihwanEom do you have plan to solve task adapt issue and ignore label issue in semi-sl cls in this PR or in this sprint?

mpa/modules/datasets/composed_dataloader.py
jaegukhyun: Got it

mpa/cls/trainer.py
supersoob: Agree with @harimkang and @jaegukhyun. I think we should find a way to adapt batch size for unlabeled dataset. And I also used unlabeled data hook for seg. I hope you go over this part. I don't think it doesn't seem good to load composedDL by checking the dataset length.

mpa/modules/datasets/composed_dataloader.py
supersoob: i.dataset is weird a little. could you rename i to loader..?

mpa/modules/hooks/model_ema_v2_hook.py
```diff
@@ -41,7 +41,10 @@ def __init__(self, ema_decay=0.9995, interval=1, start_epoch=0, dataset_len_thr=
 
     def before_train_epoch(self, runner):
         if not hasattr(self, "use_ema"):
-            self.use_ema = len(runner.data_loader.dataset) > self.dataset_len_thr
+            if isinstance(runner.data_loader.dataset, list):
```
eunwoosh: are there any chances that runner.data_loader.dataset has list type?

jaegukhyun: Jihwan add dataset attr as a list for ComposedDL in this PR

recipes/stages/_base_/data/cls_semi_data.py
supersoob: I suggest to align it by setting up the meeting. Could you summarize the agendas? I will upload PR till the meeting, so we can compare the code change. I personally think that data_pipeline doesn't have to be separated when it globally used across the models and unlabeled_hook is needed.

mpa/modules/hooks/model_ema_v2_hook.py
supersoob: Checking up if it's semi or not by type is quite strange for me. I hope we can revisit this too.

recipes/stages/_base_/data/cls_semi_data.py
jaegukhyun: @JihwanEom Sorry, for my mistake. I don't know that classification does not have default data_pipeline.py in supervised learning
@harimkang I don't want to impose use of unlabeled data hook for building unlabeled dataset and ComposeDL. However I want to unify the way to build unlabeled dataset and ComposedDL. What do you think of this?

supersoob (changes requested):
I cannot see stage.py change here for separation from incremental config. I doubt that it can cause the side effect -> loss/sampler use.. I hope you check this again.

recipes/stages/classification/semisl.yaml
JihwanEom: No, I'm not mainly responsible, only focusing on recovering semi-sl in otx. @sovrasov might be responsible for future works in multi-class semi sl.

mpa/modules/hooks/model_ema_hook.py
```diff
@@ -125,6 +125,15 @@ def _copy_model(self):
 
     def _ema_model(self):
         momentum = min(self.momentum, 1.0)
+
```
supersoob: Please remove if it's not used anymore.

mpa/cls/trainer.py
JihwanEom: These code lines are removed, please check updated code to support unlabeled batch size. Thank you.

mpa/cls/__init__.py
JihwanEom: Can I postpone merging semi-linear head and semi-nonlinear head? I felt we need a design discussion about this.

mpa/cls/trainer.py
JihwanEom: These lines are also updated, please review again. Thank you.

mpa/modules/datasets/composed_dataloader.py
JihwanEom: Removed, thank you for suggestion.

mpa/modules/hooks/model_ema_v2_hook.py
JihwanEom: Also removed by using unlabeled hook, thank you for suggestion.

JihwanEom:
Close by https://github.com/openvinotoolkit/training_extensions/pull/1473

PR #102, opened 2022-12-08
[OTX] Add SupCon recipes for classification task

arrufat:
This PR contains recipes for training models using a [SupCon](https://arxiv.org/abs/2004.11362)-like style.

These recipes aim to increase the current classification baseline when the data is scarce.

Procedure:
- Train a model with two heads:
  - classification head with standard cross-entropy loss.
  - SupCon head with custom SelfSL loss (currently it uses the [Barlow Twins](https://arxiv.org/abs/2103.03230) loss).
- At inference, the SupCon head is discarded, hence incurring no overhead at runtime.

The recipes include:
- a data pipeline using TwoCrop transform (to generate two views of an image for the SelfSL head)
- a SupConClassifier (in charge of handling the two views of images)
- a SupConClsHead (the one that uses the dual-head approach)
- a SelfSL loss (the Barlow Twins loss)

So far, this method will be used in OTX when specifying the `TrainType` as `SUPERVISEDCONTRASTIVE`.

### DIFF
--- a/mpa/modules/models/heads/supcon_cls_head.py
+++ b/mpa/modules/models/heads/supcon_cls_head.py
@@ -32,6 +32,7 @@ class SupConClsHead(BaseHead):
             topk = (topk,)
         for _topk in topk:
             assert _topk > 0, "Top-k should be larger than 0"
+        topk = (1, ) if num_classes < 5 else (1, 5)
         super().__init__(init_cfg=init_cfg)
 
         self.topk = topk
@@ -40,7 +41,7 @@ class SupConClsHead(BaseHead):
 
         # Set up the standard classification head
         self.num_classes = num_classes
-        self.linear = nn.Linear(in_features=in_channels, out_features=self.num_classes)
+        self.fc = nn.Linear(in_features=in_channels, out_features=self.num_classes)
 
         # Set up the auxiliar head
         out_channels = aux_mlp["out_channels"]
@@ -66,7 +67,7 @@ class SupConClsHead(BaseHead):
         """
 
         losses = dict(loss=0.0)
-        fc_feats = self.linear(x)
+        cls_score = self.fc(x)
 
         bsz = gt_label.shape[0]
         # make sure we have two views for each label and split them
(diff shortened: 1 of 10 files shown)


recipes/stages/classification/supcon.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+optimizer:
+    type: SGD
+    lr: 0.005
+    momentum: 0.9
+    weight_decay: 0.0005
+
+evaluation:
+    metric: ['accuracy', 'class_accuracy']
+
```
jaegukhyun, line 21: Could you explain why task adapt op disappear in this recipe?f

jaegukhyun (commented):
Genreally, LGTM. I left one comment

recipes/stages/classification/supcon.yaml
arrufat: Task adapt replaces whichever loss function I specify in the model with CrossEntropyLoss.
And I think it also changed the number of classes or something.

It wasn't working before, but maybe after the refactoring I pushed this morning, it might work.

I will check and revert it if it works.

arrufat: Yes, I've run a couple of tests, and it works now with `task_adapt` after the refactoring. Thank you!

mpa/modules/models/heads/supcon_cls_head.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, num_classes: int, in_channels: int, aux_mlp, loss, aux_loss, topk=(1,), init_cfg=None, **kwargs):
+        if in_channels <= 0:
+            raise ValueError(f"in_channels={in_channels} must be a positive integer")
+        if num_classes <= 0:
+            raise ValueError("at least one class must be exist num_classes.")
+
+        if isinstance(topk, int):
+            topk = (topk,)
+        for _topk in topk:
+            assert _topk > 0, "Top-k should be larger than 0"
```
jaegukhyun, line 34: Is it okay if _topk is larger than num_classes?

jaegukhyun (commented):
I left one comment, and I have a question. Where is the part for dividing learning rate by 2? I heard that is necessary because of changes in batch size

mpa/modules/models/heads/supcon_cls_head.py
arrufat: Fixed!

arrufat:
> I left one comment, and I have a question. Where is the part for dividing learning rate by 2? I heard that is necessary because of changes in batch size

As discussed with Wonju, for now, this logic is not implemented, we thought it would be better to do it manually and, once it because the default method for small datasets, add the logic in the code. But it's up for discussion.

jaegukhyun (approved):
LGTM, @harimkang, @supersoob, @JihwanEom anyone, could you check this PR once?

harimkang (approved):
If this works well, there seems to be no structural problem. LGTM.

PR #103, opened 2022-12-08
[OTX/MPA/Semi-SL-Seg] Enable training and evaluation for semi supervised learning in segmentation & Refactor stage.py to separate task configs

supersoob:
**This PR includes**

- New segmentor -> Mean Teacher for semi-sl
- New trainer/inferrer/stage for semi-sl
- Refactor stage.py for each task
- Custom pipeline methods for unlabeled data augmentation
- Revisit is_task_adapt flag

TODOs
- [x] Refactor v3 - pre-commit/rename
- [x] Import issue in unlabeled data hook
- [x] Check test codes and results


mpa/cls/stage.py
```diff
@@ -35,7 +35,8 @@ def configure(self, model_cfg, model_ckpt, data_cfg, training=True, **kwargs):
             else:
                 cfg.model = copy.deepcopy(model_cfg.model)
 
-        if cfg.model.pop('task', None) != 'classification':
+        cfg.model_task = cfg.model.pop('task', 'classification')
```
JihwanEom: Do you have any reason to use `cfg.model_task` instead of `cfg.task_type`?

mpa/modules/datasets/pipelines/seg_custom_pipelines.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+
+        return results
+
+    def __repr__(self):
+        return self.__class__.__name__
+
+
+@PIPELINES.register_module()
+class BranchImage(object):
```
JihwanEom: Is it hard to share `BranchImage` of detection?

mpa/modules/datasets/seg_unlabeled_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+
+from mmseg.datasets import CustomDataset
+
+logger = get_logger()
+
+
+@DATASETS.register_module()
+class UnlabeledSegDataset(CustomDataset):
+    """Dataset wrapper for Semi-SL Semantic Seg experiments.
+    Input : splits of labeled & unlabeld datasets
```
JihwanEom: `unlabeld` -> `unlabeled`

mpa/modules/datasets/seg_unlabeled_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+import numpy as np
+from mpa.utils.logger import get_logger
+
+from mmseg.datasets import CustomDataset
+
+logger = get_logger()
+
+
+@DATASETS.register_module()
+class UnlabeledSegDataset(CustomDataset):
```
JihwanEom: Do we need this dataset? I think this dataset class is not needed if we use ComposedDL. How do you think @harimkang

mpa/modules/hooks/unlabeled_data_hook.py
```diff
@@ -2,15 +2,14 @@
...
 from mmcv.runner import get_dist_info
 from mmcv.runner import HOOKS, Hook
-
-from mmdet.datasets import build_dataset, build_dataloader
-
 from mpa.modules.datasets.composed_dataloader import ComposedDL
 from mpa.utils.logger import get_logger
 
 logger = get_logger()
+task_lib_name = dict(classification="mmcls", detection="mmdet", segmentation="mmseg")
```
JihwanEom: Great work! How about moving this hook into OTX?

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+from mmseg.ops import resize
+import torch
+import functools
+from collections import OrderedDict
+from mpa.utils.logger import get_logger
+logger = get_logger()
+
+
+@SEGMENTORS.register_module()
+class MeanTeacherNaive(BaseSegmentor):
```
JihwanEom: Could you update name from `MeanTeacherNaive` -> `MeanTeacher`? `Naive` does not have ... meaningful somethings :)

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+    def __init__(self, orig_type=None, unsup_weight=0.1, warmup_start_iter=30, **kwargs):
+        logger.info('MeanTeacherNaive Segmentor init!')
+        super(MeanTeacherNaive, self).__init__()
+        self.test_cfg = kwargs['test_cfg']
+        self.warmup_start_iter = warmup_start_iter
+        self.count_iter = 0
+
+        cfg = kwargs.copy()
+        if orig_type == 'EncoderDecoder':
```
JihwanEom: We might not need `if` statement by removing OCR head

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+                                   mode='bilinear',
+                                   align_corners=self.align_corners)
+            conf_from_teacher, pl_from_teacher = torch.max(torch.softmax(teacher_logit, axis=1), axis=1, keepdim=True)
+
+        losses = dict()
+
+        x = self.model_s.extract_feat(img)
+        x_u = self.model_s.extract_feat(ul_s_img)
+        loss_decode, _ = self.model_s._decode_head_forward_train(x, img_metas, gt_semantic_seg=gt_semantic_seg)
+        loss_decode_u, _ = self.model_s._decode_head_forward_train(x_u, ul_img_metas, gt_semantic_seg=pl_from_teacher)
```
JihwanEom, line 71: As I said before, please note if strong augmented image is rotated or cropped, pseudo label from teacher will not be aligned. ex) 90 rotated image - not rotated pseudo label

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+        self.unsup_weight = unsup_weight
+
+        # Hooks for super_type transparent weight load/save
+        self._register_state_dict_hook(self.state_dict_hook)
+        self._register_load_state_dict_pre_hook(
+            functools.partial(self.load_state_dict_pre_hook, self)
+        )
+
+    def extract_feat(self, imgs):
+        return self.model_s.extract_feat(imgs)
```
JihwanEom, line 34: Could you check teacher inference? Unbiased teacher on detection uses teacher model to extract_feat and test.

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+import torch
+import functools
+from collections import OrderedDict
+from mpa.utils.logger import get_logger
+logger = get_logger()
+
+
+@SEGMENTORS.register_module()
+class MeanTeacherNaive(BaseSegmentor):
+    def __init__(self, orig_type=None, unsup_weight=0.1, warmup_start_iter=30, **kwargs):
```
JihwanEom: How about `semisl_start_iter` or `semi_start_iter` instead of `warmup_start_inter`?

mpa/seg/semisl/inferrer.py
```diff
@@ -0,0 +1,172 @@
...
+
+        cfg = self.configure(model_cfg, model_ckpt, data_cfg, training=False, **kwargs)
+        self.logger.info('infer!')
+
+        mmcv.mkdir_or_exist(osp.abspath(cfg.work_dir))
+
+        outputs = self.infer(cfg, dump_features)
+        # outputs = np.array(outputs)
+
+        # Save outputs
```
JihwanEom: Now, I think we can remove these comments :)

mpa/cls/stage.py
supersoob: Well, no specific reason, but I think task_type is more like 'instance-segmentation', 'multilabel',,...

mpa/modules/datasets/seg_unlabeled_dataset.py
supersoob: I made this for implementing cutmix method for the next task. Making unlabeled_dataset is easiest way to generate 2 images for cutmix. So I left as this.

mpa/modules/models/segmentors/mean_teacher_segmentor.py
supersoob: Semi-seg are all inferencing with student models. Does it need to match with detection task?

mpa/modules/models/segmentors/mean_teacher_segmentor.py
supersoob: Actually that step does not needed, because BranchImage is used. Please check out the code for pipelines.

mpa/modules/datasets/pipelines/seg_custom_pipelines.py
supersoob: I guess you want to inherit BranchImage from detection pipelines. But that sharing makes coupling with detection task. Plus it is not like hooks which is registered from mmcv, so I hope to leave its own BranchImage class just for seg pipeline registration.

mpa/modules/datasets/seg_unlabeled_dataset.py
JihwanEom: I don't think we need this class only for cutmix. We can implement cutmix with 1 unlabeled loader with ComposedDL.

mpa/modules/models/segmentors/mean_teacher_segmentor.py
JihwanEom: You don't have to match but it's suggestion. As far as I know, in academia, teacher model inferences when evaluate the model.

mpa/modules/datasets/seg_unlabeled_dataset.py
supersoob: I don't know I understand your point exactly, but if we impelement cutmix like you said, ComposedDL should be motified for seg cutmix, which is used across all the tasks, even other tasks does not use cutmix. I think it's better to deal it inside the dataset, not the loader.

mpa/modules/datasets/seg_unlabeled_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self, orig_type=None, **kwargs):
+        # Original dataset
+        dataset_cfg = kwargs.copy()
+        if 'cutmix' in dataset_cfg:
+            logger.warning("Currently cutmix is not supported. It will be added soon.")
+        dataset_cfg['type'] = orig_type
+        self.unlabeled_dataset = build_dataset(dataset_cfg)
+
+        self.num_unlabeled = len(self.unlabeled_dataset)
+        self.unlabeled_index = np.random.permutation(self.num_unlabeled)
```
kprokofi: Why do we need this? Obtain int -> permutate -> get the index of the permutated indexes ...

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def aug_test(self, imgs, img_metas, **kwargs):
+        return self.model_s.aug_test(imgs, img_metas, **kwargs)
+
+    def forward_dummy(self, img, **kwargs):
+        return self.model_s.forward_dummy(img, **kwargs)
+
+    def forward_train(self, img, img_metas, gt_semantic_seg, **kwargs):
+        self.count_iter += 1
+        if self.warmup_start_iter > self.count_iter or 'extra_0' not in kwargs:
```
kprokofi: What is the "extra_0"?

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+            teacher_logit = resize(input=teacher_logit,
+                                   size=ul_w_img.shape[2:],
+                                   mode='bilinear',
+                                   align_corners=self.align_corners)
+            conf_from_teacher, pl_from_teacher = torch.max(torch.softmax(teacher_logit, axis=1), axis=1, keepdim=True)
+
+        losses = dict()
+
+        x = self.model_s.extract_feat(img)
+        x_u = self.model_s.extract_feat(ul_s_img)
```
kprokofi, line 69: why do we use different images for teacher and student? `ul_t_img` and `ul_s_img`

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+        loss_decode, _ = self.model_s._decode_head_forward_train(x, img_metas, gt_semantic_seg=gt_semantic_seg)
+        loss_decode_u, _ = self.model_s._decode_head_forward_train(x_u, ul_img_metas, gt_semantic_seg=pl_from_teacher)
+
+        for (k, v) in loss_decode_u.items():
+            if v is None:
+                continue
+            losses[k] = (loss_decode[k] + loss_decode_u[k]*self.unsup_weight)
```
kprokofi, line 76: losses[k] = {k : loss_decode[k] + v*self.unsup_weight for k,v in loss_decode_u.items() is v is not None}
BTW, loss is dict, because we can have several losses?
maybe we can add unlabeled data to `_decode_head_forward_train` and modify this function?

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+        return losses
+
+    @staticmethod
+    def state_dict_hook(module, state_dict, *args, **kwargs):
+        """Redirect student model as output state_dict (teacher as auxilliary)
+        """
+        logger.info('----------------- MeanTeacherNaive.state_dict_hook() called')
+        output = OrderedDict()
+        for k, v in state_dict.items():
+            if 'model_s.' in k:
```
kprokofi: if k.startswith("model_s"):
    output[k[8:]] = v

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+        return output
+
+    @staticmethod
+    def load_state_dict_pre_hook(module, state_dict, *args, **kwargs):
+        """Redirect input state_dict to teacher model
+        """
+        logger.info('----------------- MeanTeacherNaive.load_state_dict_pre_hook() called')
+        for k in list(state_dict.keys()):
+            v = state_dict.pop(k)
+            if 'model_s.' not in k:
```
kprokofi: We disregard `model_s` in keys before saving snapshot. We may skip this checking. 
But, anyway it is better to check if k.startswith('model_s.'). What if some backbone include this name inside, not only starts with it.

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+                state_dict['model_s.'+k] = v
```
kprokofi:
```suggestion
                state_dict['model_s.' + k] = v
```

mpa/modules/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -0,0 +1,102 @@
...
+                state_dict['model_t.'+k] = v
```
kprokofi:
```suggestion
                state_dict['model_t.' + k] = v
```

recipes/stages/segmentation/semisl.py
```diff
@@ -1,49 +1,25 @@
...
 
 optimizer = dict(
-    type='SGD',
+    _delete_=True,
+    type='Adam',
     lr=1e-3,
-    momentum=0.9,
-    weight_decay=0.0005
+    eps=1e-08,
+    weight_decay=0.0
```
kprokofi, line 12: w/o weight decay??

recipes/stages/segmentation/semisl.py
```diff
@@ -1,49 +1,25 @@
...
+    '../_base_/data/seg_semisl.py',
+    '../_base_/models/segmentors/segmentor.py',
     './train.py',
-    '../_base_/models/segmentors/seg_semisl.py'
 ]
 
 optimizer = dict(
-    type='SGD',
+    _delete_=True,
+    type='Adam',
```
kprokofi, line 9: Is Adam better? Did you check it?

recipes/stages/_base_/data/seg_semisl.py
```diff
@@ -0,0 +1,37 @@
...
+    test=dict(
+        type=__dataset_type,
+        data_root=__data_root,
+        pipeline=__test_pipeline
+    ),
+    unlabeled=dict(
+        type=__dataset_type,
+        data_root=__data_root,
+        pipeline=__unlabeled_pipeline,
+        # cutmix=True
```
kprokofi: remove it for now

recipes/stages/_base_/data/seg_semisl.py
```diff
@@ -0,0 +1,37 @@
...
+    ),
+    test=dict(
+        type=__dataset_type,
+        data_root=__data_root,
+        pipeline=__test_pipeline
+    ),
+    unlabeled=dict(
+        type=__dataset_type,
+        data_root=__data_root,
+        pipeline=__unlabeled_pipeline,
```
kprokofi: Do we use different basic augmentations for unlabeled data?

recipes/stages/_base_/data/pipelines/seg_semisl.py
```diff
@@ -0,0 +1,56 @@
...
+            dict(type='Resize', keep_ratio=False),
+            dict(type='RandomFlip'),
+            dict(type='Normalize', **__img_norm_cfg),
+            dict(type='ImageToTensor', keys=['img']),
+            dict(type='Collect', keys=['img'])
+        ]
+    )
+]
+
+unlabeled_pipeline = [
```
kprokofi: There are a lot of code duplication. Could you use `append` for new augs to train_pipeline list?

mpa/seg/stage.py
```diff
@@ -162,77 +66,15 @@ def configure_task_cls_incr(self, cfg, task_adapt_type, org_model_classes, model
...
-                cfg.optimizer.lr = lr
+        if 'hyperparams' in cfg:
+            hyperparams = kwargs.get('hyperparams', None)
+            if hyperparams is not None:
+                bs = hyperparams.get('bs', None)
+                if bs is not None:
+                    cfg.data.samples_per_gpu = bs
+
+                lr = hyperparams.get('lr', None)
+                if lr is not None:
```
kprokofi, line 102: I wonder. What if `lr` is None?

kprokofi (commented):
first bunch of comments

mpa/modules/models/segmentors/mean_teacher_segmentor.py
sovrasov: I'd replace 8 with len("model_s") to avoid hardcoded numbers. "model_s" could be places into a variable, since it's used several times

mpa/seg/stage.py
```diff
@@ -21,138 +17,46 @@ def configure(self, model_cfg, model_ckpt, data_cfg, training=True, **kwargs):
...
-        # Other hyper-parameters
-        if 'hyperparams' in cfg:
-            self.configure_hyperparams(cfg, training, **kwargs)
-
-        return cfg
-
-    def configure_model(self, cfg, training, **kwargs):
-        pass
-
-    def configure_task(self, cfg, training, **kwargs):
```
kprokofi, line 94: We can't just delete all of the methods and hope that it will be reimplemented in children classes. Let's make SegStage as abstract class as we don't have intentions to use it directly, what do you think?

mpa/seg/train.py
```diff
@@ -19,6 +19,7 @@
 from mpa.seg.builder import build_dataset
 from mpa.utils.data_cpu import MMDataCPU
 
+
 def set_random_seed(seed, deterministic=False):
```
kprokofi, line 23: (I know that it is not your code, but it is something that I noticed) This function doesn't use anywhere, even more we have `set_random_seed` in Stage.py based on the config. Let's delete it.

mpa/seg/trainer.py
```diff
@@ -22,14 +22,15 @@
 from mmseg.utils import collect_env
 
 from mpa.registry import STAGES
-from mpa.seg.stage import SegStage
+from .incr.stage import IncrSegStage
```
kprokofi: You divided stages into two different training types. But now we have not intuitive structure of files. Let's move everything related to Incremental stuff to the `incr` folder. Let's keep in shared folder `seg` only shared files

mpa/modules/models/segmentors/mean_teacher_segmentor.py
kprokofi: agree. Let's avoid explicit numbers. len() will be more suitable

mpa/seg/semisl/inferrer.py
```diff
@@ -0,0 +1,172 @@
...
+from mmseg.models import build_segmentor
+from mpa.modules.hooks.recording_forward_hooks import FeatureVectorHook
+from mpa.registry import STAGES
+from .stage import SemiSegStage
+from mpa.stage import Stage
+import torch
+
+
+@STAGES.register_module()
+class SemiSegInferrer(SemiSegStage):
```
kprokofi, line 21: there is a lot of code duplication compared to seg/inferrer. I thing we may avoid it. Maybe we can try multiple inheretence? My idea to create base classes in `seg` folder. Then in `incr` and `semi` folders we may create Stage, Inferrer, Trainer. New Inferrer and Trainer can be inherited from new Stage and base Inferrer / Trainer. Concreatly we may create `SemiSegInferrer(SegInferrer, SemiSegStage)`. Or you can come up with other changes, but now I don't like current structure and code duplication.

mpa/modules/datasets/pipelines/seg_custom_pipelines.py
```diff
@@ -0,0 +1,121 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
```
kprokofi: Why this file isn't in `transforms` folder?

mpa/modules/hooks/unlabeled_data_hook.py
```diff
@@ -21,13 +20,20 @@ def __init__(
...
         **kwargs
     ):
         super().__init__(**kwargs)
+
         # Build unlabeled dataset & loader
+        m = importlib.import_module(f"{task_lib_name[task_type]}.datasets")
+        build_dataset = getattr(m, "build_dataset")
+        build_dataloader = getattr(m, "build_dataloader")
+
         logger.info('In UnlabeledDataHook.before_epoch, creating unlabeled dataset...')
```
kprokofi: it is not in `before_epoch`

mpa/modules/hooks/unlabeled_data_hook.py
```diff
@@ -44,7 +50,7 @@ def __init__(
     def before_epoch(self, runner):
```
kprokofi: Sorry, I have a question regarding this method. Why we create data loader every time before the new epoch and not once in `before_train` ?. Maybe I don't understand something

mpa/modules/datasets/seg_unlabeled_dataset.py
kprokofi: I don't understand either. All what this dataset do - create dataset like for the supervised one and iter it. What the reason of doing this? How it will help to implement cutmix?
We may build `unlabeled_dataset = build_dataset(dataset_cfg)` and use it as is. Then - mix unlabeled images in the batch during training for the cutmix.

kprokofi (commented):
Second bunch of comments

mpa/modules/datasets/pipelines/seg_custom_pipelines.py
jaegukhyun: I agreee, this should be below transforms folder

mpa/modules/models/segmentors/mean_teacher_segmentor.py
supersoob: That is the flag whether we obtain unlabeled data or not. It's from ComposedDL.

mpa/modules/models/segmentors/mean_teacher_segmentor.py
supersoob: They are same images but with different pipelines. ul_t_img -> weakly augmented image, ul_s_img -> strongly augmented image.

recipes/stages/segmentation/semisl.py
supersoob: the initial setting with SGD did not bring improvement compared to adam, which is exactly same with incremental optimizer setup. I did not conduct the thorough search for hyper parameters. But checked that this config is obviously working better than supervised. We should plan for searching optimal one later with help of advanced hpo which will be time-consuming.

recipes/stages/_base_/data/seg_semisl.py
supersoob: Yes.

mpa/modules/hooks/unlabeled_data_hook.py
supersoob: I think it's not good to move to OTX. Because it is called and defined in MPA still. We better move them to OTX after we all move stage.py and task folders.

mpa/seg/train.py
supersoob: Can I refactor the codes inside stage.py in the next PR? This PR scope is separating the stage.py config with semisl.

mpa/modules/models/segmentors/mean_teacher_segmentor.py
supersoob: I think this code implementation is more intuitive than dict comprehension.

supersoob:
@jaegukhyun @kprokofi @JihwanEom I revised codes based on your reviews. I request you to review again. I resolved your comments. If you need to add comments or still have a questions to resolved question, unresolve it or add some.

mpa/seg/semisl/inferrer.py
```diff
@@ -0,0 +1,166 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def configure(self, model_cfg, model_ckpt, data_cfg, training=False, **kwargs):
+        cfg = super().configure(model_cfg, model_ckpt, data_cfg, training=training, **kwargs)
+
+        cfg.model.type = cfg.model.orig_type
+        cfg.model.pop("orig_type", False)
+        cfg.model.pop("unsup_weight", False)
+        cfg.model.pop("warmup_start_iter", False)
```
JihwanEom: from `warmup_start_iter`to `semisl_start_iter`?

mpa/seg/semisl/trainer.py
```diff
@@ -0,0 +1,165 @@
...
+            output_ckpt_path = best_ckpt_path[0]
+        best_ckpt_path = glob.glob(os.path.join(cfg.work_dir, 'best_mIoU_*.pth'))
+        if len(best_ckpt_path) > 0:
+            output_ckpt_path = best_ckpt_path[0]
+        return dict(final_ckpt=output_ckpt_path)
+
+    @staticmethod
+    def train_worker(gpu, target_classes, datasets, cfg, distributed=False, validate=False,
+                     timestamp=None, meta=None):
+        # logger = get_logger()
```
JihwanEom: Could you remove this line?

recipes/stages/_base_/data/seg_semisl.py
```diff
@@ -0,0 +1,30 @@
+__dataset_type = ''
+__data_root = ''
+__pipeline = ''
+
+data = dict(
+    train=dict(
+        type='RepeatDataset',
```
JihwanEom, line 7: Do we need `RepeatDataset` wrapper?

supersoob: I tried to remove but the error came out.. I need to clean and go over the stage.py to fix it,(it's gonna be a time-consuming job, removing redundant cfg). Let's leave as is now and remove them later.

mpa/seg/semisl/trainer.py
supersoob: Sure, I would remove it this time, other annotations will be removed in other refactoring PRs.

mpa/seg/semisl/inferrer.py
supersoob: You found a potential bug. Thanks. Revised.

JihwanEom (approved):

recipes/stages/_base_/data/seg_semisl.py
```diff
@@ -0,0 +1,30 @@
+__dataset_type = ''
```
kprokofi, line 1: should these fields be empty?

kprokofi: Okay, I see it on the OTX side

kprokofi (approved):

PR #104, opened 2022-12-12
[OTE / XAI] Handle two stage detector in the inferrer.py

dongkwan-kim01:
This is the hot-fix while running the Geti side OTE inference, fix handling the detection saliency map hook in the `inferrer.py`.
image


goodsong81 (changes requested):
Please redirect to releases/OTE_0.4.0 branch

goodsong81:
Could you add a test to reproduce this bug? We couldn't find this in OTE CI.

dongkwan-kim01:
@goodsong81 Yes, I've changed the branch to release.

> Could you add a test to reproduce this bug? We couldn't find this in OTE CI.

Yes, seems strange why it didn't occur in the OTE test.. I'll also check and add it.

harimkang:
Thank you for your hard work!
When will this be reproduced?

mpa/det/inferrer.py
```diff
@@ -161,11 +161,21 @@ def dummy_dump_features_hook(mod, inp, out):
...
+            saliency_hook = SaliencyMapHook(eval_model.module.backbone)
+
         # Use a single gpu for testing. Set in both mm_val_dataloader and eval_model
         if is_module_wrapper(model):
             model = model.module
+
+        # Inference with hooks
         with eval_model.module.backbone.register_forward_hook(feature_vector_hook):
-            with DetSaliencyMapHook(eval_model.module) if dump_saliency_map else nullcontext() as saliency_hook:
+            with saliency_hook:
```
goodsong81: Could you explain why we could not resolve this issue in the previous PR -> https://github.com/openvinotoolkit/model_preparation_algorithm/pull/99/files ?

dongkwan-kim01:
@harimkang @goodsong81 
Hi, I'm checking the test code but currently can't find the reason why two-stage detectors are not making errors in the previous PR and passed the test codes. I'll share you when I find the reason.

dongkwan-kim01:
@harimkang @goodsong81 The reason why the test code has passed is currently `dump_saliency_map` parameter is `False` for all tests in the OTE cli side, so actually all the saliency maps are `None` when we run the ote cli(train, eval, ...)
@negvet is writing for the test [PR](https://github.com/openvinotoolkit/training_extensions/pull/1430/files) so I think it'll be resolved soon.

mpa/det/inferrer.py
dongkwan-kim01: This is my bad, should have fixed this too

harimkang:
> @harimkang @goodsong81 The reason why the test code has passed is currently `dump_saliency_map` parameter is `False` for all tests in the OTE cli side, so actually all the saliency maps are `None` when we run the ote cli(train, eval, ...) @negvet is writing for the test [PR](https://github.com/openvinotoolkit/training_extensions/pull/1430/files) so I think it'll be resolved soon.

It would be nice if tests `task.infer` of one sample with `dump_saliency_map=True` is added to the ote cli test.

goodsong81 (approved):
Changes look good to me but could you create OTE PR w/ some tests for this PR and run CI tests?

dongkwan-kim01:
@harimkang @goodsong81 Yes, I've made a [PR](https://github.com/openvinotoolkit/training_extensions/pull/1433) to check from a single sample saliency map exporting step, though it could be improved in the [test_xai improvement PR](https://github.com/openvinotoolkit/training_extensions/pull/1430/files)

negvet:
> It would be nice if tests `task.infer` of one sample with `dump_saliency_map=True` is added to the ote cli test.

`dump_saliency_map` is defined by `InferenceParameters` instance:
`dump_saliency_map = not inference_parameters.is_evaluation if inference_parameters else True`
We cannot control it from the CLI level AFAICS. Can we?

But we have some task.infer API tests https://github.com/openvinotoolkit/training_extensions/blob/46b831bcdb7db8e3dc08e819c37969de55c53b4d/external/model-preparation-algorithm/tests/api_tests/test_ote_detection_api.py#L221-L253 that actually run inference with `dump_saliency_map = True` (`inference_parameters is None`). The problem is that it is only for single stage detector `DEFAULT_DET_TEMPLATE_DIR = osp.join("configs", "detection", "mobilenetv2_atss_cls_incr")`.

To test this PR we need task.infer API tests for two-stage detectors, which we do not have running for now, as I can see. Or am I missing it?

negvet (approved):

harimkang:
> To test this PR we need task.infer API tests for two-stage detectors, which we do not have running for now, as I can see. Or am I missing it?

Yes, that's right. I think our test is missing the API test for Mask-RCNN (2 stage detector).

goodsong81:
Let's handle missing tests in another PRs. I'm merging this PR.

dongkwan-kim01:
I might be mistaken, but aren't we testing the two-stage detectors in the `test_instance_segmentation.py`?

PR #106, opened 2022-12-13
[OTX-MMCLS] Enable NNCF

cih9088:
### [Related OTX PR](https://github.com/openvinotoolkit/training_extensions/pull/1435)

**Changes**
- replace `model` parameter in MPA task train method with `model_builder` which is callable and pickable object
  - NNCF models would be built from `model_builder` function (OTX PR) given to each training method.
- output metadata for NNCF from MPA train method
- use mmcls train API
- do not send `valid_label_mask` to gpu always if it is available
- fix state dict error of `SAMImageCalssifier` if it is nested.

### DIFF
--- a/mpa/det/inferrer.py
+++ b/mpa/det/inferrer.py
@@ -37,7 +37,7 @@ class DetectionInferrer(DetectionStage):
         eval = kwargs.get('eval', False)
         dump_features = kwargs.get('dump_features', False)
         dump_saliency_map = kwargs.get('dump_saliency_map', False)
-        model = kwargs.get("model", None)
+        model_builder = kwargs.get("model_builder", None)
         if mode not in self.mode:
             return {}
 
@@ -48,7 +48,7 @@ class DetectionInferrer(DetectionStage):
 
         # mmcv.mkdir_or_exist(osp.abspath(cfg.work_dir))
 
-        outputs = self.infer(cfg, eval=eval, model=model, dump_features=dump_features,
+        outputs = self.infer(cfg, eval=eval, model_builder=model_builder, dump_features=dump_features,
                              dump_saliency_map=dump_saliency_map)
 
         # Save outputs
@@ -72,7 +72,7 @@ class DetectionInferrer(DetectionStage):
         print(json_dump)
         """
 
-    def infer(self, cfg, eval=False, model=None, dump_features=False, dump_saliency_map=False):
+    def infer(self, cfg, eval=False, model_builder=None, dump_features=False, dump_saliency_map=False):
         samples_per_gpu = cfg.data.test.pop('samples_per_gpu', 1)
         if samples_per_gpu > 1:
             # Replace 'ImageToTensor' to 'DefaultFormatBundle'
@@ -113,18 +113,20 @@ class DetectionInferrer(DetectionStage):
             target_classes = dataset.CLASSES
 
         # Model
-        if model is None:
-            cfg.model.pretrained = None
-            if cfg.model.get('neck'):
-                if isinstance(cfg.model.neck, list):
-                    for neck_cfg in cfg.model.neck:
-                        if neck_cfg.get('rfp_backbone'):
-                            if neck_cfg.rfp_backbone.get('pretrained'):
-                                neck_cfg.rfp_backbone.pretrained = None
-                elif cfg.model.neck.get('rfp_backbone'):
-                    if cfg.model.neck.rfp_backbone.get('pretrained'):
-                        cfg.model.neck.rfp_backbone.pretrained = None
+        cfg.model.pretrained = None
+        if cfg.model.get('neck'):
+            if isinstance(cfg.model.neck, list):
+                for neck_cfg in cfg.model.neck:
+                    if neck_cfg.get('rfp_backbone'):
+                        if neck_cfg.rfp_backbone.get('pretrained'):
+                            neck_cfg.rfp_backbone.pretrained = None
+            elif cfg.model.neck.get('rfp_backbone'):
+                if cfg.model.neck.rfp_backbone.get('pretrained'):
+                    cfg.model.neck.rfp_backbone.pretrained = None
 
+        if model_builder is not None:
+            model = model_builder(cfg)
+        else:
             model = build_detector(cfg.model)
         model.CLASSES = target_classes
 
--- a/mpa/det/stage.py
+++ b/mpa/det/stage.py
@@ -122,16 +122,6 @@ class DetectionStage(Stage):
                         seed=cfg.seed
                     )
                 )
-            if 'dataset' in cfg.data.train:
-                train_cfg = self.get_train_data_cfg(cfg)
-                if cfg.data.train.get('otx_dataset', None) is not None:
-                    train_cfg.otx_dataset = cfg.data.train.pop('otx_dataset')
-                if cfg.data.train.get('labels', None) is not None:
-                    train_cfg.labels = cfg.data.train.get('labels')
-                if cfg.data.train.get('data_classes', None) is not None:
-                    train_cfg.data_classes = cfg.data.train.pop('data_classes')
-                if cfg.data.train.get('new_classes', None) is not None:
-                    train_cfg.new_classes = cfg.data.train.pop('new_classes')
 
     def configure_task(self, cfg, training, **kwargs):
         """Adjust settings for task adaptation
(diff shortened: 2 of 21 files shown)


mpa/modules/hooks/checkpoint_hook.py
```diff
@@ -53,14 +53,15 @@ def after_train_epoch(self, runner):
             return
         if hasattr(runner, 'save_ckpt'):
             if runner.save_ckpt:
-                if runner.save_ema_model:
+                # TODO: Do we need this even after we moved from torchreid to mmcls?
```
sovrasov: MPA classification now also supports EMA model

mpa/modules/hooks/checkpoint_hook.py
```diff
@@ -53,14 +53,15 @@ def after_train_epoch(self, runner):
...
+                if getattr(runner, "save_ema_model", False):
                     backup_model = runner.model
                     runner.model = runner.ema_model
                 runner.logger.info(f'Saving checkpoint at {runner.epoch + 1} epochs')
                 if self.sync_buffer:
                     allreduce_params(runner.model.buffers())
                 self._save_checkpoint(runner)
-                if runner.save_ema_model:
+                if getattr(runner, "save_ema_model", False):
```
sovrasov: Normally, runner should have this argument unless someone adds a runner, which is not inherited from `EpochRunnerWithCancel`, so this workaround is not required.

mpa/modules/hooks/checkpoint_hook.py
cih9088: Removed `TODO` in https://github.com/openvinotoolkit/model_preparation_algorithm/pull/106/commits/352d48aa9a102ad2526cf2b0967fc5dc1b07f180

mpa/modules/hooks/checkpoint_hook.py
cih9088: `AccuracyAwareRunner`, a runner for `NNCF` task, is not inherited from `EpochRunnerWithCancel`.

sovrasov: Then it's a violation of LSP, the inheritance hierarchy is broken. If we have a classification-specific implementation of a NNCF task, why can't we make a subclass inherited from both this runner and AccuracyAwareRunner?

cih9088: Do you mean `*RunnerWithCancel` should be **the super classes** not `*Runner` from mmcv in OTX/MPA? If it is, I would just make `AccuracyAwareRunner` be inherited from `EpochRunnerWithCancel`.

mpa/det/stage.py
```diff
@@ -122,16 +122,6 @@ def configure_data(self, cfg, training, **kwargs):
...
-            if 'dataset' in cfg.data.train:
-                train_cfg = self.get_train_data_cfg(cfg)
-                if cfg.data.train.get('otx_dataset', None) is not None:
-                    train_cfg.otx_dataset = cfg.data.train.pop('otx_dataset')
-                if cfg.data.train.get('labels', None) is not None:
-                    train_cfg.labels = cfg.data.train.get('labels')
-                if cfg.data.train.get('data_classes', None) is not None:
-                    train_cfg.data_classes = cfg.data.train.pop('data_classes')
-                if cfg.data.train.get('new_classes', None) is not None:
-                    train_cfg.new_classes = cfg.data.train.pop('new_classes')
```
goodsong81, line 134: Is this block not needed eventually?

goodsong81 (approved):
Looks good to me, thank you.

mpa/det/stage.py
cih9088: I believe so. OTX stuff should be handled in OTX side.

Here is a simple flow.

In OTX, `otx_dataset`, `labels` attributes would be initialized to `None` in [this function](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmcls/otx/algorithms/detection/adapters/mmdet/utils/config_utils.py#L248-L249) properly [depending on data configuration recipe provided (`self._recipe_cfg.data`)](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmcls/otx/algorithms/detection/adapters/mmdet/utils/config_utils.py#L245) when [initializing recipe](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmcls/otx/algorithms/detection/tasks/inference.py#L259). After that [this function](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmcls/otx/algorithms/common/adapters/mmcv/utils.py#L312-L332) aligns actual data configuration structure (`self._data_cfg.data`) to the initialized recipe's one (`self._recipe_cfg.data`) while [initializing task](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmcls/otx/algorithms/common/tasks/training_base.py#L250-L251).
`data_classes`, `new_classes` attributes would be properly defined in [here](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmcls/otx/algorithms/common/tasks/training_base.py#L109-L112) as long as data configuration of recipe does not have dataset wrappers of [theses kinds](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/dataset_wrappers.py#L17).
Then in MPA, `data_cfg` could be merged into `recipe_cfg` without any miss match concerns.

mpa/modules/hooks/checkpoint_hook.py
sovrasov: Yes, that's it. All cls runners are supposed to implement functionality from `EpochRunnerWithCancel`

cih9088: Fixed in https://github.com/openvinotoolkit/model_preparation_algorithm/pull/106/commits/1bee390495bce892270de1be3bc286b6c42fce85

sovrasov (approved):

mpa/det/inferrer.py
```diff
@@ -113,18 +113,20 @@ def infer(self, cfg, eval=False, model=None, dump_features=False, dump_saliency_
...
+        cfg.model.pretrained = None
+        if cfg.model.get('neck'):
+            if isinstance(cfg.model.neck, list):
+                for neck_cfg in cfg.model.neck:
+                    if neck_cfg.get('rfp_backbone'):
+                        if neck_cfg.rfp_backbone.get('pretrained'):
+                            neck_cfg.rfp_backbone.pretrained = None
+            elif cfg.model.neck.get('rfp_backbone'):
+                if cfg.model.neck.rfp_backbone.get('pretrained'):
+                    cfg.model.neck.rfp_backbone.pretrained = None
```
wonjuleee, line 125: when model doesn't contain neck, `cfg.model.pretrained` is going to be `None`. Is this okay?

mpa/modules/models/heads/custom_multi_label_non_linear_cls_head.py
```diff
@@ -117,7 +118,6 @@ def get_valid_label_mask(self, img_metas):
             mask = torch.Tensor([1 for _ in range(self.num_classes)])
             if 'ignored_labels' in meta and meta['ignored_labels']:
                 mask[meta['ignored_labels']] = 0
-            mask = mask.cuda() if torch.cuda.is_available() else mask
             valid_label_mask.append(mask)
```
wonjuleee, line 121: shouldn't this be `to(x.device)` here?

mpa/seg/inferrer.py
```diff
@@ -95,18 +95,20 @@ def infer(self, cfg, model=None, dump_features=False):
...
-                if isinstance(cfg.model.neck, list):
-                    for neck_cfg in cfg.model.neck:
-                        if neck_cfg.get('rfp_backbone'):
-                            if neck_cfg.rfp_backbone.get('pretrained'):
-                                neck_cfg.rfp_backbone.pretrained = None
-                elif cfg.model.neck.get('rfp_backbone'):
-                    if cfg.model.neck.rfp_backbone.get('pretrained'):
-                        cfg.model.neck.rfp_backbone.pretrained = None
-            cfg.model.test_cfg.return_repr_vector = True
+        cfg.model.pretrained = None
```
wonjuleee, line 98: Here also.

mpa/modules/models/heads/custom_multi_label_non_linear_cls_head.py
cih9088: This method doesn't get any tensors as its arguments so that we can't. Instead, it would be moved to proper device in line # 98.

mpa/det/inferrer.py
cih9088: To be honest, I have no clue. This part is here as it is.

mpa/seg/inferrer.py
cih9088: To be honest, I have no clue. This part is here as it is.

PR #107, opened 2022-12-13
[OTE / XAI][Develop] Handle two stage detector in the inferrer.py

dongkwan-kim01:
This is the hot-fix while running the Geti side OTE inference, fix handling the detection saliency map hook in the `inferrer.py`.
image


dongkwan-kim01:
@goodsong81 Sorry but this is the same as #104. Could you also approve this?

goodsong81 (approved):

PR #108, opened 2022-12-13
[OTX] Add warmstart for classification recipe

sungchul1:
This PR includes
* Warmstart for classification recipe
* Disable validation during warmstart
* Enable LARS in OTX

FYI.
Warmstart experiment for reproducibility was conducted with input size=32 for CIFAR10 and CIFAR100 as in previous experiments, but input size in OTX is 224 with OTX default setting (https://github.com/openvinotoolkit/model_preparation_algorithm/blob/otx/recipes/stages/_base_/data/pipelines/semisl_pipeline.py#L3).
So, input size in this recipe is set to 224 to match OTX unlike this experiment.
We can discuss how to set up this config.


mpa/cls/trainer.py
```diff
@@ -117,16 +117,18 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
         # cfg.dump(osp.join(cfg.work_dir, 'config.yaml')) # FIXME bug to save
         # logger.info(f'Config:\n{cfg.pretty_text}')
 
+        # TODO (sungchul): check other condition for vaildation
+        validate = False if cfg.model.type in ['BYOL'] else True
```
jaegukhyun: Could you explain why BYOL has different value with other algorithms?

recipes/stages/_base_/data/warmstart_cls_data.py
```diff
@@ -0,0 +1,83 @@
...
+    dict(type='RandomAppliedTrans',
+         transforms=[dict(type='Solarization')],
+         p=0.2),
+    dict(type='Normalize',  **__img_norm_cfg),
+    dict(type='ImageToTensor', keys=['img']),
+    dict(type='Collect', keys=['img'])
+]
+
+data = dict(
+    samples_per_gpu=128,
```
jaegukhyun: Does warmstart requires specific batch size? 
I think this will be replaced by otx side.

recipes/stages/classification/warmstart.yaml
```diff
@@ -0,0 +1,45 @@
...
+    '../_base_/dist/dist.py',
+    '../_base_/logs/tensorboard_logger.py',
+    '../_base_/optimizers/lars.py',
+    '../_base_/runners/iter_runner.py',
+    '../_base_/schedules/cos_anneal.py',
+    '../_base_/data/warmstart_cls_data.py',
+]
+
+# in default.py settings
+cudnn_benchmark: true
```
jaegukhyun: I think cudnn and deterministic is set for experimental purpose, but this should follow other recipes

jaegukhyun (changes requested):
Generally, LGTM. I left some comments

mpa/cls/trainer.py
sungchul1: Most models not using warmstart probably have `SAMImageClassifier` as `model.type`, but warmstart methods have `BYOL`.
This condition is just for distinguishing warmstart methods (which don't use validation) from other models.
If you have better condition or it is required to modify config, please let me know.

recipes/stages/_base_/data/warmstart_cls_data.py
sungchul1: Exactly. 256 is set in OTX and it will be replaced, and `workers_per_gpu`, too.
Is it required to change it to another value?

recipes/stages/classification/warmstart.yaml
sungchul1: I'll fix it.

sungchul1: Actually, I wanted to avoid `default.py` because I don't want to use `task_adapt` in warmstart, so I removed `default.py` in `_base_` and added others in `default.py` except for `task_adapt`.
In `default.py`, `cudnn_benchmark` is set to `True` as default setting, and it seems not for experimental purpose unlike you said. (Of course, I agree `deterministic` and it will be set to `False`.)

mpa/cls/trainer.py
jaegukhyun: Than, my question is "Why BYOL algo do not take validation phase?"

recipes/stages/classification/warmstart.yaml
jaegukhyun: If cudnn_benchmark is True in default.py then here, it should be True

mpa/cls/trainer.py
sungchul1: Warmstart conducts self-supervised learning during small epochs with target dataset to adapt model weights to target dataset.
For self-supervised learning, validation is not required, so validation set is not set in OTX.
You can refer https://github.com/openvinotoolkit/training_extensions/pull/1438/files#diff-3b8d1f587e0a91a74d0d736158cc1d7d76a706871315620662cb4e00afdfd7beR148-R155.

sungchul1: if cfg.data.val is None

sungchul1:
I updated validation condition as we discussed.
@goodsong81 @jaegukhyun Could you check this update?

wonjuleee (approved):
LGTM

jaegukhyun (approved):

sungchul1:
@wonjuleee @jaegukhyun 
Thanks for your efforts!
Experiments with input size = 224 are in progress.
If there are config changes after the experiments, I'll update them and merge this PR.

sungchul1:
Update:
* batch size : 256 → 64 (due to vram memory issue)
* rebase

PR #112, opened 2022-12-16
[OTX] Fix Classification Stage & Classifier for supporting public backbone

harimkang:
## Summary
BACKBONE Registy change mmcls to mmcv used in Classification Stage
If the output of the backbone is a tuple (other tasks backbone), 
adjust the classification head to receive the output of the last layer

Through this work, i confirmed the backbone replacement, build, and training of mmdet, mmseg, torchvision, and omz based on SAMImageClassifier

image

### DIFF
--- a/mpa/modules/models/classifiers/sam_classifier.py
+++ b/mpa/modules/models/classifiers/sam_classifier.py
@@ -257,16 +257,18 @@ class SAMImageClassifier(ImageClassifier):
            Overriding for OpenVINO export with features
         """
         x = self.backbone(img)
-        # For Global Backbones (det/seg/etc..),
-        # In case of tuple, only the feat of the last layer is used.
-        if isinstance(x, tuple):
-            x = x[-1]
 
         if torch.onnx.is_in_onnx_export():
             self.featuremap = x
 
         if self.with_neck:
             x = self.neck(x)
+
+        # For Global Backbones (det/seg/etc..),
+        # In case of tuple, only the feat of the last layer is used.
+        if isinstance(x, tuple):
+            x = x[-1]
+
         return x
 
     def simple_test(self, img, img_metas):
--- a/mpa/modules/models/classifiers/sam_classifier.py
+++ b/mpa/modules/models/classifiers/sam_classifier.py
@@ -257,6 +257,11 @@ class SAMImageClassifier(ImageClassifier):
            Overriding for OpenVINO export with features
         """
         x = self.backbone(img)
+        # For Global Backbones (det/seg/etc..),
+        # In case of tuple, only the feat of the last layer is used.
+        if isinstance(x, tuple):
+            x = x[-1]
+
         if torch.onnx.is_in_onnx_export():
             self.featuremap = x
 
--- a/mpa/modules/models/classifiers/sam_classifier.py
+++ b/mpa/modules/models/classifiers/sam_classifier.py
@@ -257,17 +257,16 @@ class SAMImageClassifier(ImageClassifier):
            Overriding for OpenVINO export with features
         """
         x = self.backbone(img)
+        # For Global Backbones (det/seg/etc..),
+        # In case of tuple or list, only the feat of the last layer is used.
+        if isinstance(x, (tuple, list)):
+            x = x[-1]
 
         if torch.onnx.is_in_onnx_export():
             self.featuremap = x
 
         if self.with_neck:
             x = self.neck(x)
-
-        # For Global Backbones (det/seg/etc..),
-        # In case of tuple, only the feat of the last layer is used.
-        if isinstance(x, tuple):
-            x = x[-1]
 
         return x
 
--- a/mpa/cls/stage.py
+++ b/mpa/cls/stage.py
@@ -121,7 +121,7 @@ class ClsStage(Stage):
             return
 
         # update model layer's in/out configuration
-        from mmcls.models.builder import BACKBONES as backbone_reg
+        from mmcv.cnn import MODELS as backbone_reg
         layer = build_from_cfg(cfg.model.backbone, backbone_reg)
         layer.eval()
         input_shape = [3, 224, 224]


mpa/modules/models/classifiers/sam_classifier.py
```diff
@@ -257,11 +257,17 @@ def extract_feat(self, img):
            Overriding for OpenVINO export with features
         """
         x = self.backbone(img)
+        # For Global Backbones (det/seg/etc..),
+        # In case of tuple or list, only the feat of the last layer is used.
+        if isinstance(x, (tuple, list)):
+            x = x[-1]
```
harimkang, line 263: The original classification head can only receive `Tensors`, and cannot receive `tuples` or `lists`.
However, backbones used in some detection or segmentation give the output of multiple layers in the form of a tuple or list. So, I first set the output of the last layer among the outputs emitted by the backbone to be used.

if this doesn't feel right for you, or if you have any new ideas, let me know.

harimkang:
latest otx(feature/otx) + this mpa branch
- otx cli test for classification
image
- otx api test for classification
image

JihwanEom (approved):

harimkang:
@sungmanc @goodsong81 This only applied to the current `SAMImageClassifier`. Should it apply to other Classifiers as well? (All currently used classification models use `SAMImageClassifier`) What do you think?

mpa/modules/models/classifiers/sam_classifier.py
supersoob: I wanna catch up why det and seg backbone affects to classifier in classification. Is it a case that like loading the backbone weight previously trained in the way of det and seg to classification task?

harimkang: The main purpose of public backbone support is to provide usable backbone support regardless of the task.
The model used in det/seg is configured to receive the output of multiple layers in the form of a list or tuple in backbone forward. However, classification is composed of the output of the last layer as a target, so cannot receive a list or tuple.
That's why we need to do this.

And I think the answer to the second question is 'yes'. The backbone can be trained independently of any task, and making these weights usable is also a goal. (So far I haven't tested this..)

supersoob: It's like a puzzling the models extending the range across tasks. Codes for considering the other task which is registered classification task was kinda confused me first. Now reasonable. Thanks for your kind explanation.

supersoob (approved):

mpa/modules/models/classifiers/sam_classifier.py
harimkang, line 263: Hi! @negvet I need your advice so leave a comment.. :)
I have implemented the current public backbone (detection or segmentation) for use by attaching it to the classification model. -> Ex: {mmdet backbone + classifier}
The detection backbone basically emits `multiple outputs` and returns a `tuple or list`, and in order to use it for classification, the last output is subtracted and passed in this way now. `x = x[-1]`

However, recently, as the `ActivationMapHook` on the MPA side was changed to `ReciproCAMHook`, there was a problem in the infer process (Only `classification` + when the output of backbone is a `tuple or list` -> I think this currently has no test code.)
https://github.com/openvinotoolkit/model_preparation_algorithm/blob/c1e801f24595a95594be3520ecdf97efdd73263d/mpa/modules/hooks/recording_forward_hooks.py#L266 
I think the conflicting cause part is here, and when the output is a `tuple or list`, the `feature_map` at `index 0` (default `fpn_idx=0`) is taken out and used. 

It seems that there is a conflict in the calculation because the head is different from the known size, and it has been confirmed that it works when `fpn_idx=-1`

I'm not sure if my explanation is clear, but summary is that there's a problem when the `feature_map` exported from backbone is a `tuple or list` pass to classification head, and it seems to be caused by different indices. (My guess so far is that...)

Can you give me an idea or solution to this problem or what I am doing wrong or what needs to be fixed?


P.S: @goodsong81 @sungmanc If you have any advice please!

negvet: Hi Harim

Thanks for a notice!

You are right.

I do not know why `fpn_idx` defaults to 0. It was 0 before and I did not update it (my bad, and yes, IMO we do not have tests for num_backbone_outputs > 1 for cls). `fpn_idx` is the index of the backbone output (while increasing the index - it approaching the head). By default we should take the last layer output for explanation purposes. So `fpn_idx` should default to -1. 

`fpn_idx=-1` is a good default for `ActivationMapHook`. But this is not a must to be. In practice we might want to see `fpn_idx=0` for example (`ActivationMapHook` will work with any `fpn_idx` from possible idx range). So we should keep this argument. 

`ReciproCAMHook`, on the other hand, process the output from the backbone and forward it through the cls head. Then it makes a big difference which `fpn_idx` we pick (we need to make sure that cls head can get it as input). 
Assuming that the cls heads only works with the higher fpn level (`fpn_idx=-1`), then for `ReciproCAMHook` we also need to assign `fpn_idx=-1`. 

> Assuming that the cls heads only works with the higher fpn level

If this is true, then we actually need to hardcode it to be `x = x[-1]`. Because no other options are available (in contrast to `ActivationMapHook`).

Can assume that there might be other advanced options: (1) cls head work with NOT the higher fpn level, (2) cls head can take as input all fpn levels. These more advanced use cases are not supported by `ReciproCAMHook`. It is a good to raise an error in this case.

Hope it was clear.

Should I open a PR to update `fpn_idx` default values or you have it already somewhere?

sungmanc: @harimkang @negvet , In this situation, we should select `x=x[-1]` for the classifier and `fpn_idx=-1` for the ReciproCAMHook. And actually, IMO, ReciproCAMHook can only take the last backbone layer because it needs to connect with `cls_head` like as below.
image

And I think changing the cls_head (advanced option 1, 2) is inappropriate. Since CAM assumes that all layers included in the model were pre-trained, changing the size of cls_head might not work as we anticipated. 
(i.e. Assume that the model was trained on 1,024 size cls_head and we want to use a 2,048 feature map. In this case, we can change the size of cls_head to 1,024 --> 2,048 to resolve mis-size issue. However, in this case, cls_head will not work because it was trained on 1,024)

So, let's use `fpn_idx=-1` as a solution. And maybe we need to extract the saved tensor and avoid re-forward to make the tensor output.

PR #121, opened 2022-12-22
[Hot-fix][OTX] Resolve Instance segmentation model config issue in detection explainer

harimkang:
## Summary
This PR addresses the following issues that occur during instance segmentation tests.
image

- Instance segmentation otx cli test
image
- Detection otx cli test
image

### DIFF
--- a/mpa/det/explainer.py
+++ b/mpa/det/explainer.py
@@ -75,7 +75,22 @@ class DetectionExplainer(DetectionStage):
 
         # Target classes
         target_classes = explain_dataset.CLASSES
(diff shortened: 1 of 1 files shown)


eunwoosh (approved):
LGTM. But there is same code in OTX side. when mpa is combined with otx, then it's better to make this part as separate function and reuse it.

harimkang:
> LGTM. But there is same code in OTX side. when mpa is combined with otx, then it's better to make this part as separate function and reuse it.

i agree with you, thanks!

jaegukhyun (approved):

supersoob (approved):

PR #129, opened 2023-01-31
Fix ignore filtering in hierarchical heads

sovrasov:
Batch size after filtering should be different then the initial size if there are filtered elements

### DIFF
--- a/mpa/modules/models/heads/custom_hierarchical_linear_cls_head.py
+++ b/mpa/modules/models/heads/custom_hierarchical_linear_cls_head.py
@@ -95,12 +95,13 @@ class CustomHierarchicalLinearClsHead(MultiLabelClsHead):
         if self.compute_multilabel_loss:
             head_gt = gt_label[:, self.hierarchical_info['num_multiclass_heads']:]
             head_logits = cls_score[:, self.hierarchical_info['num_single_label_classes']:]
-            valid_mask = head_gt >= 0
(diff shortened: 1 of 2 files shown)


mpa/modules/models/heads/custom_hierarchical_linear_cls_head.py
```diff
@@ -95,12 +95,13 @@ def forward_train(self, x, gt_label, **kwargs):
         if self.compute_multilabel_loss:
             head_gt = gt_label[:, self.hierarchical_info['num_multiclass_heads']:]
             head_logits = cls_score[:, self.hierarchical_info['num_single_label_classes']:]
-            valid_mask = head_gt >= 0
-            head_gt = head_gt[valid_mask].view(*valid_mask.shape)
-            head_logits = head_logits[valid_mask].view(*valid_mask.shape)
+            valid_batch_mask = head_gt >= 0
+            head_gt = head_gt[valid_batch_mask, ]
```
goodsong81, line 99: Does this mean head_gt[valid_batch_mask, :] ?

goodsong81 (approved):
LTGM, but changed the target to release branch.

mpa/modules/models/heads/custom_hierarchical_linear_cls_head.py
sovrasov: Yes, it's the same

